PR #149, opened 2020-05-08
Add redis conn close func to client

t-asaka:
I added function to client to close redis connection


hibiken:
Great, thank you for adding this!

PR #205, opened 2020-10-14
Add tls support to command line tool

strobus:
This PR adds a `--tls_server` argument to the CLI tool. Provide the name of a server that is included in the redis server or cluster's certificate. Note, the `tls.Config` struct only allows one server to be added for validation see [crypto/tls](https://golang.org/pkg/crypto/tls/#Config). A future enhancement could provide additional arguments for specifying client certificates, etc., though such use cases may be infrequent for redis.

Note: this has been tested with cluster mode. Thank you for adding that!

### DIFF
--- a/tools/asynq/cmd/root.go
+++ b/tools/asynq/cmd/root.go
@@ -5,6 +5,7 @@
 package cmd
 
 import (
+	"crypto/tls"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


hibiken:
Great! Thank you for adding TLS support 🎉 
PR LGTM 👍

PR #207, opened 2020-11-07
Add Unregister method to Scheduler

### LINKED ISSUE #204: [FEATURE REQUEST] UnRegister function to unregister task entries
New Scheduler functionality is quite awaited feature. Maybe you would consider following functionalities:

- `RemoveEntry` or `UnRegister` function to remove entry by `entryID`. So instead of stopping cron and manually removing the entries maybe it could be done with a function
- View Entries by time period. For e.g. Entries for `@every 30s`, Entries for `@daily`

Also this will allow possibility to handle CRON via REST or any other services

hibiken:
Closes #204

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `Unregister` method is added to `Scheduler` to remove a registered entry.
+
 ## [0.13.0] - 2020-10-13
 
 ### Added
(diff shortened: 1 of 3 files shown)


hibiken:
Closing this pr for now since I don't see the use case for this.

JordyMoos:
Hi @hibiken ,

We are using your library and are in need of this functionality. As we are adding tasks dynamically and we want our non developer colleagues to be able to change the cronspec and remove the registered task independently. The GUI for this is made outside of Asynq, but now we need to be able to unregister tasks for these two scenarios:

- We want the cronspec to change for a specific registered task
- We want to remove the cron all together for a specific registered task (for example our business entity gets deleted)

Please let me know if you can revive this PR for us or if you want some more information and or help.

Best Regards

PR #219, opened 2020-12-19
fix: resolve go vet lint

mingrammer:


hibiken:
Thank you for fixing!

PR #235, opened 2021-01-29
add support for custom redis logger

bp0lr:
Hello @hibiken, 
first of all, let me thank you for your great work :)

I'm using termui and asynq to simulate a worker and all the logs are redirected to logs files.
I don't know why, but go-redis print some logs direct to stdout / stderr.
because of this, when I got a redis error, my UI get broke.

imagen

go-redis comes with a function to define a global logger.
So I have added a wrapper function on asynq to be able to set that logger.

I did my test using the last version of asynq and I'm using logrus for logging.

To make some, just need to initialize a logrus instance and past it to the function.
```
cacheLogger := stdlog.New(jlog.Log.Writer(), "", 0)	
asynq.SetReditLogger(cacheLogger)
```

and the result was:
imagen

this will also open an option to set a custom logger just for redis errors.
Maybe the logger could be part of the configuration like asynq does on server creation. I don't know.
Let me know if you have a better solution.

thanks!

### DIFF
--- a/asynq.go
+++ b/asynq.go
@@ -11,6 +11,8 @@ import (
 	"strconv"
 	"strings"
 
+	stdlog "log"
+
 	"github.com/go-redis/redis/v7"
 )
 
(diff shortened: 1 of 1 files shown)


hibiken:
@bp0lr Thank you for reporting this issue! 

I think I've seen go-redis log messages while I was working on asynq, but didn't look into it further. So thank you for bringing this to my attention.

That said I don't want to make an API change at this time because of this dependency's quirk. I'll keep this issue open as a reminder to address this in the future. But in the meantime, is it possible to call `redis.SetLogger(logger)` within your package to have redis package output logs to a log file?

bp0lr:
after spare some more time doing tests, you were right, It could be called from my project.

thank you!

hibiken:
ok great! I'll close this pr, but thank you for reporting the issue. I'll revisit if this becomes a common issue for other users.

PR #251, opened 2021-03-11
Add Unregister method to Scheduler

### LINKED ISSUE #204: [FEATURE REQUEST] UnRegister function to unregister task entries

hibiken:
Closes #204

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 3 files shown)


JordyMoos:
Thanks for your efforts @hibiken !

PR #252, opened 2021-03-12
Refactor redis key and store messages in protobuf

### LINKED ISSUE #245: Use protobuf to encode messages stored in Redis
Currently, all task messages are serialized in JSON format and stored in redis. 
This comes with two issues:

- JSON takes more space compared to other serialization scheme (e.g. protobuf, msgpack). We can expect less memory usage with these more compact serialization schemes.
- Protocol Buffer has clearer forward/backward compatibility semantics as we evolve the library. We can expect that fields will be added/removed from the task messages stored in redis.

hibiken:
Closes #245


disc:
@hibiken Hey, when could we expect next release with protobuf support?

PR #254, opened 2021-03-20
Replaced blocking KEYS command to non-blocking SCAN

disc:
I've noticed that [asynqmon](https://github.com/hibiken/asynqmon) makes `KEYS queuename:*` redis calls every a few seconds. But [KEYS](https://redis.io/commands/keys) redis command is blocking and could affect production redis server.
I suggest to replace KEYS to [SCAN](https://redis.io/commands/scan) command - it will be more safest way to iterate all keys and get their memory usage.

More details: https://redis.io/commands/KEYS

> Warning: consider KEYS as a command that should only be used in production environments with extreme care. It may ruin performance when it is executed against large databases. This command is intended for debugging and special operations, such as changing your keyspace layout. Don't use KEYS in your regular application code. If you're looking for a way to find keys in a subset of your keyspace, consider using SCAN or sets.

### DIFF
--- a/internal/rdb/inspect.go
+++ b/internal/rdb/inspect.go
@@ -172,9 +172,17 @@ func (r *RDB) CurrentStats(qname string) (*Stats, error) {
 }
 
 func (r *RDB) memoryUsage(qname string) (int64, error) {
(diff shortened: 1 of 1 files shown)


hibiken:
@disc Thank you for spotting and fixing this!

I'll release a new version with this fix.

PR #268, opened 2021-04-28
Update README with a new gopher icon

koddr:
Hi,

I think, this gopher from Go's website favicon would be better 😉 

Task Queue Diagram

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Task queues are used as a mechanism to distribute work across multiple machines.
 
 **Example use case**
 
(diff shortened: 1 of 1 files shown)


hibiken:
@koddr Thank you for updating this!

Is it possible to remove the background color of the gophers? and change the border color to match the other shapes in the image?

koddr:
@hibiken like this?

Desktop - 1

hibiken:
Actually I like the first one! Let me merge this  🎉 
Thank you so much @koddr 🙏

Issue #293, opened 2021-07-15
[FEATURE REQUEST] Upgrade go-redis dependency to v8


strobus:
**Is your feature request related to a problem? Please describe.**
I'm ready to upgrade my projects to go-redis/v8 but asynq is holding me back. I'd rather not have 2 versions of go-redis in my stack. Are there plans to upgrade?

**Describe the solution you'd like**
Upgrade Asynq to go-redis/v8

**Describe alternatives you've considered**
N/A

**Additional context**
go-redis/v8 has been out of beta for a while.

hibiken:
@strobus Thanks for opening this issue!

Yes, I've been meaning to update to v8. I just haven't had a chance to do that yet since it was not a high priority.
I'll probably get to it within the coming week or so, but if this is urgent for you, please feel free to open a PR 👍

PR #298, opened 2021-07-20
Upgrade go-redis/redis to version 8

strobus:
This PR upgrade the `go-redis/redis/v7` dependency to `go-redis/redis/v8`. The major change between the two versions is that all methods that interact with the redis server have a `context.Context` parameter added. These context parameters have been surfaced up to the exported `asynq` interfaces.

Edit: resolves issue https://github.com/hibiken/asynq/issues/293


strobus:
TODO: I tested with a local redis docker container and all was well. However my redis cluster is in AWS (Elasticache). There were several test failures around date/time stamps being slightly off or task scores being off by one or two. My guess is these are due to a test design which requires a specific setup, probably locally. @hibiken do you have a recommended cluster setup for testing locally?

client.go
```diff
@@ -265,7 +266,7 @@ func (c *Client) Close() error {
 // By deafult, max retry is set to 25 and timeout is set to 30 minutes.
 //
 // If no ProcessAt or ProcessIn options are provided, the task will be pending immediately.
-func (c *Client) Enqueue(task *Task, opts ...Option) (*TaskInfo, error) {
+func (c *Client) Enqueue(ctx context.Context, task *Task, opts ...Option) (*TaskInfo, error) {
```
hibiken, line 269: I'm not against it, but I'm not sure if we want to make this API change in this PR. 
Unless you have a strong argument for this, I'll make internal changes without changing the public API of the package.

internal/base/base.go
```diff
@@ -637,21 +637,21 @@ func (c *Cancelations) Get(id string) (fn context.CancelFunc, ok bool) {
...
+	Schedule(ctx context.Context, msg *TaskMessage, processAt time.Time) error
+	ScheduleUnique(ctx context.Context, msg *TaskMessage, processAt time.Time, ttl time.Duration) error
+	Retry(ctx context.Context, msg *TaskMessage, processAt time.Time, errMsg string) error
+	Archive(ctx context.Context, msg *TaskMessage, errMsg string) error
+	ForwardIfReady(ctx context.Context, qnames ...string) error
+	ListDeadlineExceeded(ctx context.Context, deadline time.Time, qnames ...string) ([]*TaskMessage, error)
+	WriteServerState(ctx context.Context, info *ServerInfo, workers []*WorkerInfo, ttl time.Duration) error
+	ClearServerState(ctx context.Context, host string, pid int, serverID string) error
+	CancelationPubSub(ctx context.Context) (*redis.PubSub, error) // TODO: Need to decouple from redis to support other brokers
+	PublishCancelation(ctx context.Context, id string) error
```
hibiken, line 655: I don't think we need to change the interface Broker unless we need to.

scheduler.go
```diff
@@ -130,15 +131,15 @@ func (j *enqueueJob) Run() {
...
+	err = j.rdb.RecordSchedulerEnqueueEvent(context.Background(), j.id.String(), event)
 	if err != nil {
 		j.logger.Errorf("scheduler could not record enqueue event of enqueued task %+v: %v", j.task, err)
 	}
 }
 
 // Register registers a task to be enqueued on the given schedule specified by the cronspec.
 // It returns an ID of the newly registered entry.
-func (s *Scheduler) Register(cronspec string, task *Task, opts ...Option) (entryID string, err error) {
+func (s *Scheduler) Register(ctx context.Context, cronspec string, task *Task, opts ...Option) (entryID string, err error) {
```
hibiken, line 142: Same here. I'd wait to see if we need to make this API change.

hibiken (changes requested):
Thank you for working on this!

Do you have a reason for changing public API in this PR? Do you think users of the package will use context for deadlines, cancelation, etc?

I think we can focus just on upgrading the redis library version in this PR, so that changes are only scoped to `internal/rdb` package and shouldn't affect rest of the codebase. 
What do you think?

strobus:
I'm not opposed to keeping the exported API as is. I'm not as familiar with the internal operation of the module, so wasn't sure if we can make assumptions about the context we pass to the underlying redis library. Would we just pass `context.Background()` then?

crossworth:
Probably we should just pass `context.Background()` and if one day we need a context version we could do the approach that `database/sql` uses, create a `EnqueueContext` and `RegisterContext` or with options `client.Enqueue(task, asynq.Context(ctx))`.

see:
https://cs.opensource.google/go/go/+/refs/tags/go1.16.6:src/database/sql/sql.go;l=1548
https://github.com/go-redis/redis/issues/1694

strobus:
Yes I like those approaches, which we can add down the road. I'll back out the exported API changes in the next couple days and switch to using `context.Background()` internally.

hibiken (changes requested):
@strobus thank you for working one this. 
One last change request:
I prefer to avoid creating a globally scoped variable `ctx` (both in tests and application code).
Would you mind changing it so that the scope of the variable is limited to a method or a redis command invocation? (you can just call `context.Background()` or `context.TODO()` at the call site)

strobus:
I am happy to do it. This project is a crucial component in what I'm working on, so thank you for developing this.

internal/asynqtest/asynqtest.go
```diff
@@ -165,12 +166,12 @@ func FlushDB(tb testing.TB, r redis.UniversalClient) {
...
 	switch r := r.(type) {
 	case *redis.Client:
-		if err := r.FlushDB().Err(); err != nil {
+		if err := r.FlushDB(context.Background()).Err(); err != nil {
 			tb.Fatal(err)
 		}
 	case *redis.ClusterClient:
-		err := r.ForEachMaster(func(c *redis.Client) error {
-			if err := c.FlushAll().Err(); err != nil {
+		err := r.ForEachMaster(context.Background(), func(ctx2 context.Context, c *redis.Client) error {
```
hibiken, line 173: s/ctx2/ctx

tools/go.mod
```diff
@@ -3,21 +3,13 @@ module github.com/hibiken/asynq/tools
...
-	github.com/coreos/go-etcd v2.0.0+incompatible // indirect
-	github.com/cpuguy83/go-md2man v1.0.10 // indirect
 	github.com/fatih/color v1.9.0
-	github.com/go-redis/redis/v7 v7.4.0
-	github.com/golang/protobuf v1.4.1 // indirect
+	github.com/go-redis/redis/v8 v8.11.2
 	github.com/google/uuid v1.2.0
 	github.com/hibiken/asynq v0.17.1
 	github.com/mitchellh/go-homedir v1.1.0
-	github.com/spf13/cast v1.3.1
```
hibiken, line 15: I don't think this dependency should be removed. Would you mind running:

```sh
$ cd ./tools
$ go build ./...
```

hibiken (changes requested):
Sorry for going back and forth so many times (hopefully this is the last time!)

I added two comments. please take a look!
Also, it looks like this branch needs to be rebased on master.

tools/go.mod
strobus: This completed successfully and did not result in any changes to go.mod or go.sum. I tried under both Go 1.15 and 1.16. Note that in Go 1.16 `go build` [no longer modifies](https://golang.org/doc/go1.16#go-command) go.mod or go.sum. 

> Build commands like go build and go test no longer modify go.mod and go.sum by default. Instead, they report an error if a module requirement or checksum needs to be added or updated (as if the -mod=readonly flag were used). Module requirements and sums may be adjusted with go mod tidy or go get.

hibiken: I see. I think I was wrong there. Please ignore this comment. But please fix the other comment and then we can merge this PR 👍

strobus:
Ah yes, missed that.

hibiken:
@strobus Thank you for the hard work and patience to address all the comments 🎉 Merging the PR now.

Issue #297, opened 2021-07-20
[FEATURE REQUEST] empty taskname should not be publish


victor-wzq:
i push a task with empty taskname,
but i can't comsume this task,
i think function Enqueue should check taskname

hibiken:
@victor-wzq thank you for creating this issue! I haven't thought about this edge case so thank you for pointing this out  👍 

I think we have a couple of options:
1. `NewTask` function can panic given a empty typename 
2. `Enqueue` function can return an error if the task has an empty typename
3.  Allow tasks to have an empty typename

Let me know what you think! /cc @crossworth

crossworth:
I think we should go with option 2.

1. We should avoid panic's, we dont know where the typename could came from (maybe user input or dynamic generated) and panicing makes it harder to the caller to handle it, we should avoid panic as much as possible as a library. By returning an error we can effectively tell the caller what went wrong.

2. I think is the best option, Enqueue already returns errors like `ErrDuplicateTask`, it would be easy to handle.

3. I dont think its a good option, can cause a lot of confusion, makes debuging harder and most of the times the task should have a typename.

crossworth:
@hibiken this can be closed.

hibiken:
Closing this. Thank you @Luqqk!

PR #303, opened 2021-08-01
client.Enqueue - prevent empty task typename

Luqqk:
Hi @hibiken! Thanks for this library and sharing your work 👍 

When I was trying to get familiar with the code base and test out some of the features locally I also decided to contribute and help a little bit. I picked #297 as something that appears easy to start with. Please let me know if I missed something out or the solution should be improved/adjusted. Keep in mind that currently you're still able to queue and consume tasks with typenames like `" "` (just a single whitespace). This can be a feature but maybe also should be handled differently?

Thanks again and I'm looking forward to contribute in future!

### DIFF
--- a/client.go
+++ b/client.go
@@ -6,6 +6,7 @@ package asynq
 
 import (
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 
@@ -178,8 +179,8 @@ func (d processInOption) Value() interface{} { return time.Duration(d) }
 // ErrDuplicateTask error only applies to tasks enqueued with a Unique option.
 var ErrDuplicateTask = errors.New("task already exists")
 
-// ErrEmptyTypeTask indicates that task's typename is not specified.
-var ErrEmptyTypeTask = errors.New("task typename not specified")
+// ErrEmptyTypename indicates that task's typename is empty.
+var ErrEmptyTypename = errors.New("task typename cannot be empty")
 
 type option struct {
 	retry     int
@@ -269,8 +270,8 @@ func (c *Client) Close() error {
 //
 // If no ProcessAt or ProcessIn options are provided, the task will be pending immediately.
 func (c *Client) Enqueue(task *Task, opts ...Option) (*TaskInfo, error) {
-	if task.Type() == "" {
-		return nil, fmt.Errorf("%w", ErrEmptyTypeTask)
+	if strings.TrimSpace(task.Type()) == "" {
+		return nil, fmt.Errorf("%w", ErrEmptyTypename)
 	}
 	c.mu.Lock()
 	if defaults, ok := c.opts[task.Type()]; ok {
--- a/client.go
+++ b/client.go
@@ -178,6 +178,9 @@ func (d processInOption) Value() interface{} { return time.Duration(d) }
 // ErrDuplicateTask error only applies to tasks enqueued with a Unique option.
 var ErrDuplicateTask = errors.New("task already exists")
 
+// ErrEmptyTypeTask indicates that task's typename is not specified.
+var ErrEmptyTypeTask = errors.New("task typename not specified")
+
 type option struct {
 	retry     int
 	queue     string
@@ -266,6 +269,9 @@ func (c *Client) Close() error {
 //
 // If no ProcessAt or ProcessIn options are provided, the task will be pending immediately.
 func (c *Client) Enqueue(task *Task, opts ...Option) (*TaskInfo, error) {
+	if task.Type() == "" {
+		return nil, fmt.Errorf("%w", ErrEmptyTypeTask)
+	}
 	c.mu.Lock()
 	if defaults, ok := c.opts[task.Type()]; ok {
 		opts = append(defaults, opts...)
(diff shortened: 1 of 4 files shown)


crossworth:
Hello @Luqqk , great work!

If you don't mind I would like to make a few comments.

> Keep in mind that currently you're still able to queue and consume tasks with typenames like " " (just a single whitespace). This can be a feature but maybe also should be handled differently?

Currently on the handler registration the check is with ` == ""` as well, so maybe this is not a problem, but maybe we should use `TrimSpace` in the future, let's see what @hibiken thinks about it.
https://github.com/hibiken/asynq/blob/73f930313cef4434c7a62f54bece8051788fa89c/servemux.go#L101

Maybe the error name should be:  `ErrEmptyTypename` just to make it more precise.

I have one more consideration, maybe the error should be handled on the `asynq/internal/rdb/rdb.go` as well, since `ErrDuplicateTask` is handled in a similar way, but I dont see why right now, since rdb is internal and the only client is `client.go`.

Luqqk:
Thanks for quick feedback @crossworth!

I'll adjust the error name as you suggested. 

> I have one more consideration, maybe the error should be handled on the asynq/internal/rdb/rdb.go as well, since ErrDuplicateTask is handled in a similar way, but I dont see why right now, since rdb is internal and the only client is client.go.

Yup, I'm not sure about this one as well and that's the reason I added it only in `client.go` for now.

Let's wait for @hibiken, maybe he'll shed some light on it.

client.go
```diff
@@ -178,6 +178,9 @@ func (d processInOption) Value() interface{} { return time.Duration(d) }
 // ErrDuplicateTask error only applies to tasks enqueued with a Unique option.
 var ErrDuplicateTask = errors.New("task already exists")
 
+// ErrEmptyTypeTask indicates that task's typename is not specified.
+var ErrEmptyTypeTask = errors.New("task typename not specified")
```
hibiken, line 182: I don't expect many users encountering this error, so surfacing a sentinel error value via public API seems like an overkill. Let's not add this to package API yet. We can always add this to public API if needed :)

Also, another nit-pick: can we change the error message to `"task typename cannot be empty"` (reason: Technically user has *specified* typename, it's just empty)

client.go
```diff
@@ -266,6 +269,9 @@ func (c *Client) Close() error {
 //
 // If no ProcessAt or ProcessIn options are provided, the task will be pending immediately.
 func (c *Client) Enqueue(task *Task, opts ...Option) (*TaskInfo, error) {
+	if task.Type() == "" {
```
hibiken, line 272: Like @crossworth suggested, let's `TrimSpace` here so that we can check for any blank strings.
Please make sure to update corresponding check in `servemux.go` as well!

hibiken (changes requested):
Thank you for the contribution @Luqqk!

@crossworth thanks for reviewing it so quickly, I really appreciate your comment as always :)

Let's defer exposing sentinel error value for this case, I expect very few users will encounter this error anyways.

client.go
Luqqk: Thanks for the feedback. I fully agree that the error message you suggested sounds better. Adjusted it as requested.

client.go
Luqqk: Added `TrimSpace` in `client.go` and `servemux.go`

crossworth (approved):

client.go
```diff
@@ -266,6 +270,9 @@ func (c *Client) Close() error {
 //
 // If no ProcessAt or ProcessIn options are provided, the task will be pending immediately.
 func (c *Client) Enqueue(task *Task, opts ...Option) (*TaskInfo, error) {
+	if strings.TrimSpace(task.Type()) == "" {
+		return nil, fmt.Errorf("%w", ErrEmptyTypename)
```
hibiken, line 274: this should be sufficient, since we don't want to export a sentinel error value for this case.
```go
return nil, fmt.Errorf("task typename cannot be empty")
```

client.go
```diff
@@ -178,6 +179,9 @@ func (d processInOption) Value() interface{} { return time.Duration(d) }
 // ErrDuplicateTask error only applies to tasks enqueued with a Unique option.
 var ErrDuplicateTask = errors.New("task already exists")
 
+// ErrEmptyTypename indicates that task's typename is empty.
+var ErrEmptyTypename = errors.New("task typename cannot be empty")
```
hibiken, line 183: Like I said in the last comment, I don't think we should export this error value from package API (I don't expect many users encountering this error anyway).

client_test.go
```diff
@@ -585,6 +585,11 @@ func TestClientEnqueueError(t *testing.T) {
 				Queue(""),
 			},
 		},
+		{
+			desc: "With empty task typename",
+			task: NewTask("", h.JSON(map[string]interface{}{})),
+			opts: []Option{},
+		},
```
hibiken, line 592: another case for blank line?

hibiken (changes requested):
Thanks for fixing so quickly. I left a few comments, please address those as well.
Also, I forgot in my last review, but please add an entry in CHANGELOG.md file.

client.go
Luqqk: Ok, now I see what you meant. Thanks for clarifying.

client_test.go
Luqqk: Added another test case.

Luqqk:
I adjusted the code as requested. Also added the changelog entry. Small suggestion - would be nice to add an info in the "contribution guide" that changelog entry should be a part of the PR ;)

hibiken (approved):

PR #333, opened 2021-10-04
Introduce Task Results

### LINKED ISSUE #265: [FEATURE REQUEST] Processed job's payload and result
Asynq currently removes the successfully processed jobs from Redis. 
That means the information about the processed jobs also vanishes like;

- When did the client pushed the job into the queue
- When is the job completed. 
- Is a given job id processed successfully.

It would be nice to keep processed jobs until a given TTL and return a basic struct that contains something like ` id, queued_at, completed_at `.
I think to achieve that, the first thing is to generate a unique id for each task and keep them like `asynq:{default}:processed:{job_id}`

[Python RQ](https://python-rq.org/) or [Laravel Horizon](https://laravel.com/docs/8.x/horizon) might be good examples to take a look at.

hibiken:
Closes #265


asynq.go
```diff
@@ -440,3 +481,28 @@ func parseRedisSentinelURI(u *url.URL) (RedisConnOpt, error) {
...
+	select {
+	case <-w.ctx.Done():
+		return 0, fmt.Errorf("failed to result task result: %v", w.ctx.Err())
+	default:
+	}
+	return w.broker.WriteResult(w.qname, w.id, data)
+}
+
+// TaskID returns the ID of the task the ResultWriter is associated with.
+func (w *ResultWriter) TaskID() string {
```
hibiken, line 506: Not sure if we want this method. Maybe remove it for now

PR #336, opened 2021-10-27
Add counting semaphore to control distributed max concurrency

### LINKED ISSUE #328: [FEATURE REQUEST] Distributed concurrency support per queue / taskType
N/A

Is it possible to set concurrency individually for every queue like it is supported [here](https://github.com/gocraft/work#job-concurrency).

Currently I tried to wire up the server in a way where I calculate total concurrency and set max priority and assuming that the combination will provide some sort of similar effects as in `gocraft/work`. Storing the maxConcurrency and priority info per queue and using it for initialisation of the server when starting the adapter (a simple wrapper over asynq).

```go
type Adapter struct {
	taskConfigMap map[string]taskConfig
	tcMutex       sync.RWMutex
}

type taskConfig struct {
	priority       int
	maxConcurrency int
}

func (a *Adapter) server() *asynq.Server {
	return asynq.NewServer(a.redisConnOpt, asynq.Config{
		Concurrency:    a.concurrency(),
		Queues:         a.queuePriorityMap(),
	})
}

func (a *Adapter) queuePriorityMap() map[string]int {
	a.tcMutex.RLock()
	defer a.tcMutex.RUnlock()

	m := make(map[string]int)

	for tn, cfg := range f.taskConfigMap {
		if cfg.priority == 0 {
			m[tn] = 1
			continue
		}

		m[tn] = cfg.priority
	}

	return m
}

func (a *Adapter) concurrency() int {
	a.tcMutex.RLock()
	defer a.tcMutex.RUnlock()

	concurrency := 0
	for _, cfg := range f.taskConfigMap {
		if cfg.maxConcurrency == 0 {
			concurrency++
			continue
		}

		concurrency += cfg.maxConcurrency
	}

	if rtcpu := runtime.NumCPU(); concurrency < rtcpu {
		return rtcpu
	}

	return concurrency
}
```

**Additional context**
Ref: Priority Sampler from [gocraft/work](https://github.com/gocraft/work/blob/5959e69ad211c5ca37ffdf3ede02e35a5ae41d98/priority_sampler.go#L21)

ajatprabha:
Closes #328


x/rate/semaphore.go
```diff
@@ -0,0 +1,72 @@
...
+// - Returns true, iff semaphore key exists and current value is less than maxConcurrency
+// - Returns false otherwise
+func (s *Semaphore) Acquire(ctx context.Context) bool {
+	val, _ := acquireCmd.Run(ctx, s.rc, []string{semaphoreKey(s.name)}, []interface{}{s.maxConcurrency}...).Bool()
+	return val
+}
+
+// Release will release the lock on the counting semaphore.
+func (s *Semaphore) Release(ctx context.Context) {
+	s.rc.Decr(ctx, semaphoreKey(s.name))
```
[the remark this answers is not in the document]
hibiken: Fair point. I think we can make an internal context package. Just move the existing `context.go` to a separate internal package 👍

x/rate/semaphore_test.go
```diff
@@ -0,0 +1,331 @@
...
+			defer sema.Close()
+
+			for i := 0; i < len(tt.taskIDs); i++ {
+				ctx, cancel := tt.ctxFunc(tt.taskIDs[i])
+
+				got, err := sema.Acquire(ctx)
+				if got != tt.want[i] {
+					t.Errorf("%s;\nSemaphore.Acquire(ctx) returned %v, want %v", tt.desc, got, tt.want[i])
+				}
+				if (tt.wantErr == "" && err != nil) || (tt.wantErr != "" && (err == nil || err.Error() != tt.wantErr)) {
```
hibiken, line 158: This is very hard to read. Please move the error case to a separate test

x/rate/semaphore_test.go
```diff
@@ -0,0 +1,331 @@
...
+}
+
+func TestNewSemaphore_Acquire_StaleLock(t *testing.T) {
+	opt := getRedisConnOpt(t)
+	rc := opt.MakeRedisClient().(redis.UniversalClient)
+	defer rc.Close()
+
+	taskID := uuid.New()
+
+	rc.ZAdd(context.Background(), semaphoreKey("stale-lock"), &redis.Z{
```
hibiken, line 175: nit: Add some comments to describe what this set up is doing

x/rate/semaphore_test.go
```diff
@@ -0,0 +1,331 @@
...
+func TestNewSemaphore_Release(t *testing.T) {
+	testID := uuid.New()
+
+	tests := []struct {
+		desc      string
+		name      string
+		taskIDs   []uuid.UUID
+		ctxFunc   func(uuid.UUID) (context.Context, context.CancelFunc)
+		wantCount int64
+		wantErr   string
```
hibiken, line 208: Same as above. Please create a separate test function for error case.

hibiken (commented):
I think we are almost there. Just left a few comments :) Let me know if you have thoughts/questions

x/rate/semaphore.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	b, err := acquireCmd.Run(ctx, s.rc,
+		[]string{semaphoreKey(s.scope)},
+		s.maxTokens,
+		time.Now().Unix(),
+		d.Unix(),
+		taskID,
+	).Bool()
+	if err == redis.Nil {
```
hibiken, line 85: when would this happen? Could you add a comment for why we treat this differently?

x/rate/semaphore.go
```diff
@@ -0,0 +1,118 @@
+package rate
```
hibiken, line 1: Please add a package comment.

hibiken (commented):
LGTM. Just left final few nit pick comments. After those are resolved, I'll merge this PR 👍

x/rate/semaphore.go
ajatprabha: There was an issue with lua script, I missed that! It required me to return true/false values as string.

PR #370, opened 2021-12-27
fix: missing import statement in example code

chronark:
Just a quick fix.

I noticed a few imports are missing in the example code in the README


hibiken:
Thank you!

PR #375, opened 2022-01-02
feat: allow configuration of intervals

### LINKED ISSUE #359: [FEATURE REQUEST] forwarder.interval support dynamic adjustment
https://github.com/hibiken/asynq/blob/9f2c321e9804901f06b58e82f9e0ce4125d80b74/server.go#L369

mdibaiee:
fixes #359

### DIFF
--- a/server.go
+++ b/server.go
@@ -144,6 +144,12 @@ type Config struct {
 	//
 	// If unset or zero, the interval is set to 15 seconds.
 	HealthCheckInterval time.Duration
+
+	// DelayedTaskCheckInterval specifies the interval between checks run on 'scheduled' and 'retry'
+	// tasks, and forwarding them to 'pending' state if they are ready to be processed.
+	//
+	// If unset or zero, the interval is set to 5 seconds.
+	DelayedTaskCheckInterval time.Duration
 }
 
 // An ErrorHandler handles an error occured during task processing.
@@ -287,6 +293,8 @@ const (
 	defaultShutdownTimeout = 8 * time.Second
 
 	defaultHealthCheckInterval = 15 * time.Second
+
+	defaultDelayedTaskCheckInterval = 5 * time.Second
 )
 
 // NewServer returns a new Server given a redis connection option
@@ -362,11 +370,15 @@ func NewServer(r RedisConnOpt, cfg Config) *Server {
 		starting:       starting,
 		finished:       finished,
 	})
+	delayedTaskCheckInterval := cfg.DelayedTaskCheckInterval
+	if delayedTaskCheckInterval == 0 {
+		delayedTaskCheckInterval = defaultDelayedTaskCheckInterval
+	}
 	forwarder := newForwarder(forwarderParams{
 		logger:   logger,
 		broker:   rdb,
 		queues:   qnames,
-		interval: 5 * time.Second,
+		interval: delayedTaskCheckInterval,
 	})
 	subscriber := newSubscriber(subscriberParams{
 		logger:       logger,
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+vendor
 # Binaries for programs and plugins
 *.exe
 *.exe~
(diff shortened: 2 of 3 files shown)


hibiken (changes requested):
Thank you for working on this!

I think (at least for now) we only need to allow configuration of "forwarder" interval through public API. 

Also, since the name "forwarder" is very much an internal detail of the package, I think we should come up with a more descriptive name.

### Background context
The forwarder goroutine is running indefinitely and periodically checking both the *scheduled* and *retry* tasks and forwarding them to *pending* state if they are ready to be processed. The interval is currently hard-coded to 5s but some users may want to customize it. Even though we allow configuration by the user, we may want to restrict it (e.g. minimum of 1s) since we don't want to be checking this every 10ms for example.

We should also document this clearly so that it shows up in the godoc, and it's clear to the user what this configuration does.

mdibaiee:
Hey @hibiken 
Thanks for the feedback.

I worked on multiple ones since I saw your comment on the [issue](https://github.com/hibiken/asynq/issues/359):
> There're a few intervals like this one that we can make configurable. I'll try to include this change in the next release 👍

Is it a bad idea to add them all as configurable?

Regarding naming, I think it comes down to whether we make forwarder meaningful or not: we could just add the description you added as part of the documentation of the parameter and people will understand it. I find it hard to come up with a name that's self-explanatory for this case, while still being short enough. What do you think? We can do the same for all other interval configurations: explaing what exactly is the process that runs on that interval.

hibiken:
I probably made that comment without thinking too much on it :)

I think we should be conservative when making public API changes, once you add a configuration option it's hard to remove it (e.g. requires major version update after the v1 release). I don't see much use for other intervals tweaked by the users, right now.

As for the name, I was thinking in the lines of `DelayedTaskCheckInterval` (or something simliar), since both the *scheduled* and *retry* tasks are delayed tasks (i.e. tasks that will be processed in the future). The name forwarder is something I'm not super happy with so I'd rather not expose it  :)

Let me know if you have feedback/thoughts!

mdibaiee:
@hibiken okay, that makes sense 👍

server.go
```diff
@@ -144,6 +144,12 @@ type Config struct {
 	//
 	// If unset or zero, the interval is set to 15 seconds.
 	HealthCheckInterval time.Duration
+
+	// DelayedTaskCheckInterval specifies the interval between checks run on scheduled and 'retry'
```
hibiken, line 148: nit: quote the word "scheduled" to be consistent.

server.go
```diff
@@ -349,24 +365,28 @@ func NewServer(r RedisConnOpt, cfg Config) *Server {
 	syncer := newSyncer(syncerParams{
 		logger:     logger,
 		requestsCh: syncCh,
-		interval:   5 * time.Second,
+		interval:   defaultSyncInterval,
```
hibiken, line 368: revert this change.

server.go
```diff
@@ -349,24 +365,28 @@ func NewServer(r RedisConnOpt, cfg Config) *Server {
...
 	})
 	heartbeater := newHeartbeater(heartbeaterParams{
 		logger:         logger,
 		broker:         rdb,
-		interval:       5 * time.Second,
+		interval:       defaultHeartbeatInterval,
```
hibiken, line 373: revert this change.

server.go
```diff
@@ -394,7 +414,7 @@ func NewServer(r RedisConnOpt, cfg Config) *Server {
 		retryDelayFunc: delayFunc,
 		isFailureFunc:  isFailureFunc,
 		queues:         qnames,
-		interval:       1 * time.Minute,
+		interval:       defaultRecovererPollInterval,
```
hibiken, line 417: revert this change.

server.go
```diff
@@ -406,7 +426,7 @@ func NewServer(r RedisConnOpt, cfg Config) *Server {
 		logger:   logger,
 		broker:   rdb,
 		queues:   qnames,
-		interval: 8 * time.Second,
+		interval: defaultJanitorInterval,
```
hibiken, line 429: revert this change.

hibiken (commented):
Thanks for making the change quickly, I'll approve once all outstanding comments have been resolved 👍

PR #406, opened 2022-02-24
Enable rediss url parsing support

eleboucher:


hibiken (commented):
Thank you for adding support for this!

Change LGTM. Would you mind adding an entry in the CHANGELOG under "Unreleased" section?

hibiken (commented):
Sorry one more thing (forgot to mention in the last review). 
Would you mind updating the godoc for the `ParseRedisURI` function?

Issue #369, opened 2021-12-26
[BUG] Unique tasks before TTL should not be run


gmhafiz:
**Describe the bug**
The second of two identical tasks with same UUID and payload should not be sent to Redis before TTL, when the message sent could have been from a retry.

This is a related to issue https://github.com/hibiken/asynq/issues/275 with the difference that in 275, repeated tasks are allowed to be processed before TTL because the previous task is deleted. In my case, asynq must respect TTL even when the task has completed.

**To Reproduce**
Steps to reproduce the behavior (Code snippets if applicable):
1. An external API sends an HTTP request to my asynq microservice through an endpoint. 
2. This external API does a retry for 3 times, sleeping 10 seconds in between.
3. The first request could have been delayed for more than 10 seconds, prompting a first retry (second identical request).
4. That second request reaches asynq before original first request, starts processing, and completed in one second.
5. The original first request finally reaches asynq and starts processing - even though it is identical to the one that has been processed and before its TTL.

Using 
```go
info, err := u.redis.EnqueueContext(ctx, task, asynq.Unique(5*time.Minute))
```

**Expected behavior**
 - Unique() respects TTL of completed tasks
 - the original first request (which is second chronologically) should have been rejected because it is identical to the first retry (second request) and before 5 minutes 

**Environment (please complete the following information):**
 - OS: Linux
 - Version of `asynq` package v0.20.0

**Additional context**

Another way to achieve distributed lock is I manually do the following.
 - include a UUID on every request
 - extract UUID from request
 - check if UUID key is present in Redis
   - If not present, create a new key/value entry
  
```go
 client.Set(r.Context(), requestID, true, 0).Err()
```
   - If present, reject
 - Do not delete the UUID key from Redis even after the task has been completed. This is to ensure retry requests fail.
 - Once task has been completed, update the key with TTL
```go
redis.Set(ctx, requestID, true, 5*time.Minute).Err()
```

 - Once TTL has passed, Redis automatically deletes the key, allowing identical request to be processed.

Additionally, I could delete the UUID key if a task is explicitly deleted (not completed).

The other way I can think of is to use retention.
```go
asynqClient.EnqueueContext(ctx, task, asynq.TaskID(request.UUID), asynq.Retention(5*time.Minute))
```
This way, the UUID is still kept in Redis for another 5 minutes after the task is completed. This prevents the Http retry request from being accepted.

hibiken:
@gmhafiz Thank you for reporting an issue!

[...]

PR #407, opened 2022-02-24
add UniqueKey option to extend Unique's behaviour

benjajaja:
Instead of (queue, type, payload) use any custom key. This allows to
enforce unique tasks not limited to same payload and/or type.

Tested with `--redis_cluster` flag.

### DIFF
--- a/client.go
+++ b/client.go
@@ -44,6 +44,7 @@ const (
 	TimeoutOpt
 	DeadlineOpt
 	UniqueOpt
+	UniqueKeyOpt
 	ProcessAtOpt
 	ProcessInOpt
 	TaskIDOpt
@@ -70,6 +71,7 @@ type (
 	timeoutOption   time.Duration
 	deadlineOption  time.Time
 	uniqueOption    time.Duration
+	uniqueKeyOption string
 	processAtOption time.Time
 	processInOption time.Duration
 	retentionOption time.Duration
@@ -158,6 +160,15 @@ func (ttl uniqueOption) String() string     { return fmt.Sprintf("Unique(%v)", t
 func (ttl uniqueOption) Type() OptionType   { return UniqueOpt }
 func (ttl uniqueOption) Value() interface{} { return time.Duration(ttl) }
 
+// UniqueKey returns an option to set the unique key of the task, instead of the default Unique behavior.
+func UniqueKey(key string) Option {
+	return uniqueKeyOption(key)
+}
+
+func (key uniqueKeyOption) String() string     { return fmt.Sprintf("UniqueKey(%q)", string(key)) }
+func (key uniqueKeyOption) Type() OptionType   { return UniqueKeyOpt }
+func (key uniqueKeyOption) Value() interface{} { return key }
+
 // ProcessAt returns an option to specify when to process the given task.
 //
 // If there's a conflicting ProcessIn option, the last option passed to Enqueue overrides the others.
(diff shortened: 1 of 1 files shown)


hibiken:
@benjajaja Thanks for opening a pull request!

If you are able to generate a unique key, then the `TaskID` option may be sufficient for your use case :) Please take a look at [this wiki](https://github.com/hibiken/asynq/wiki/Unique-Tasks#using-taskid-option). And let me know if this doesn't work for your use case 👍

benjajaja:
@hibiken `TaskID` indeed works for uniqueness, unless you also use retention - then a completed task will block tasks of same `TaskID` until retention expires.

So that's my use case: I want some retention, so that I can query the task for a bit after completion, but I also want to use uniqueness options. `Unique` on its own isn't sufficient, because my payload has something like `userID, someOtherUniquePerTaskThingId`. This means my tasks are always "unique".

benjajaja:
@hibiken right now I'm using the `TaskID` to ensure a single task of given criteria at a time. But then I can't add retention, because the task will block even though it's completed. I want the retention so that I can properly check if a task has been completed, beyond "not existing".

benjajaja:
BTW I started by trying to add a "key" option to `Unique`, however it got awkward when it would be the only `option` wrapping two values.

hibiken:
Presumably you want to generate a `UniqueKey` string with some TTL, correct?
My previous comment was suggesting that you could generate a unique task ID that captures these two values.

For example, if you want to use a unique_key of (user_id) and TTL of 1 hour then you could generate a unique task ID in a format `<user_id>_<yyyy-mm-dd-hh:00:00>` (e.g. `"user123_2022-02-28-15:00:00"`). 

This way, all tasks fall under the same time window will be deduplicated, but a task in the next time window won't be blocked by the previous task even if that previous task is still in the queue as a completed task (actually it doesn't matter which state the previous task is in).

benjajaja:
> Presumably you want to generate a `UniqueKey` string with some TTL, correct? My previous comment was suggesting that you could generate a unique task ID that captures these two values.

No, `Unique` already has its TTL, which I only care about to prevent locking some type of task forever.

I do want retention, so that I can call `GetTaskInfo` on a completed task (otherwise this returns "not found" after completion). Retention does not play well with uniqueness-by-task-id, as it blocks new incoming tasks.

> For example, if you want to use a unique_key of (user_id) and TTL of 1 hour then you could generate a unique task ID in a format `<user_id>_<yyyy-mm-dd-hh:00:00>` (e.g. `"user123_2022-02-28-15:00:00"`).
> 
> This way, all tasks fall under the same time window will be deduplicated, but a task in the next time window won't be blocked by the previous task even if that previous task is still in the queue as a completed task (actually it doesn't matter which state the previous task is in).

A time window is interesting, but I am not bound by any time windows. It would also allow for duplicate tasks, if the tasks run longer than the window. Thank you for coming up with this, but it won't really fit my case.

BTW, as I said I am using unique task-ids so far and it works, it's just that I can't retain them for more than some seconds, as that block newer tasks. It's not a big deal, but it would be nice to retain them longer, to do some more analysis.

hibiken:
I see. Thank you for providing more details.

I've seen similar feedback/discussion around `Unique` option such as #369 and #376. 

It seems that we need to allow user to specify:
- Whether the uniqueness lock should be released on task completion
- Unique key used for the uniqueness lock

in addition to the TTL value.
I'll take this feedback and come up with a solution that addresses these issues together.

benjajaja:
Ok, thanks! I'll keep an eye on it.

PR #408, opened 2022-02-25
Add options for the clients to set default timeout and max retries

eleboucher:


hibiken:
@eleboucher thanks for opening a pull request! 
We used to have something very similar to this on the client (`Client.SetDefaultOptions`) but got removed in [v0.19.0](https://github.com/hibiken/asynq/releases/tag/v0.19.0). 

The reason for removing is to simplify the override rules for the options. 
User can now set the options when instantiating the task with `NewTask(typename string, payload []byte, opts ...Option)` Or at enqueue time with `Client.Enqueue(task *Task, opts ...Option)`, where options at enqueue time can override the options set when creating the task object. Having another place to set an option (e.g. as client default) adds another rule to this override logic and I didn't want to over complicate that. 

If you really need to have this, I suppose a user can write a simple wrapper around `asynq.Client` to set the default options at client level? Let me know your thoughts on this 👍

eleboucher:
Thanks for the explanation @hibiken it makes sense.

PR #432, opened 2022-04-10
Add Task Aggregation feature

### LINKED ISSUE #339: [FEATURE REQUEST] Group tasks by a specific key
I have a service that that sends emails to users and sometimes it may happen that there will be >20 emails of the same type to send. Not the same email this is the key difference why I can't use the unique task option. Let me give you an example of such situation:

There's an email to accept an invitation to edit file A, B, C, ... They all come from different places in the system, so the queue will hold 3 emails (the same type `id-send-invitation`) but with different data (filename: `A`, ...). I want to detect such state and get these 3 emails from queue to send only one email: `You have an invitation to many files (A,B,...) go to dashboard and accept them)

There are a few that comes to my mind:
1. allow to create queues in a runtime like `email:invitation:id:reciver@email.com` and allow to flush them every X seconds. In such a case, the task handler receives `[]Task` slice and can implement a logic what to do with each task. 
2. allow to pass a `group method` which will tell a queue how to group tasks into buckets. It could look like:
```go
func group(t *asyncq.Task) (bucket string, windowToWait time.Time){
  return t.payload.id + t.payload.email, time.Minute
}
```

Use some other services/tools/message brokers to achieve that

**Additional context**
-

hibiken:
Closes #339


krhubert:
@hibiken I will test it and use it right away! Thanks a lot, you're great!

hibiken:
@krhubert Thanks!

This is a brand new feature, so it may have some rough edges. Any feedback/questions are appreciated 👍

PR #433, opened 2022-04-13
Correct the error message to cancel an active tasks

eleboucher:


hibiken:
Thanks for catching this!

PR #448, opened 2022-05-05
docs: correct typo (deafult => default)

jeffreylo:


hibiken:
@jeffreylo Thank you!

Issue #468, opened 2022-05-17
[FEATURE REQUEST] Expose taskID when occurring Duplicated Error


linhbkhn95:
**Is your feature request related to a problem? Please describe.**
- Currently, `asynq` only common error `Duplicated TaskID` when the created task is duplicated. So, If we use uniqueID via generating like `blockNumber` then we difficult to detect taskID which is duplicate.

**Describe the solution you'd like**
Currently,` asynq` uses error const as like below
```
// ErrTaskIDConflict indicates that the given task could not be enqueued since its task ID already exists.
//
// ErrTaskIDConflict error only applies to tasks enqueued with a TaskID option.
var ErrTaskIDConflict = errors.New("task ID conflicts with another task")

```
I think we can use error struct to catching taskID.

```
// ErrTaskIdConflict indicates that another task with the same task ID already exist.
type TaskIDConflictError struct {
	id string
}

func NewTaskIDConflictError(id string) error {
	return TaskIDConflictError{
		id: id,
	}
}

func (t TaskIDConflictError) Error() string {
	return fmt.Sprintf("task id conflicts with another task at taskID=%s", t.id)
}

hibiken:
@linhbkhn95 thank you for creating an issue.

If you are getting `ErrTaskIDConflict` error, you are providing a custom ID using `TaskID` option. IMO it's redundant to include the ID in the error message, since user just called `Enqueue` with the ID.

linhbkhn95:
I think, almost case is duplicated ID from using a custom ID, So wrapping the ID into an error will provide insight to the user. Moreover, If users don't use a custom ID, sometimes, they use an ID generated for tracing. So, display ID duplicated is useful.

PR #470, opened 2022-05-17
chore(*): Expose task id which is duplicated

linhbkhn95:
Why did we need it?
- Show taskID to console when taskID is duplicated
- Solved #468


hibiken:
Thank you for creating a PR!

Unfortunately, I don't see a need for this (see my comment [here](https://github.com/hibiken/asynq/issues/468#issuecomment-1129532090))

Closing this.

Issue #456, opened 2022-05-08
[FEATURE REQUEST] Is there any way to do some specific action before periodic task enqueue?


ChihSeanHsu:
**Is your feature request related to a problem? Please describe.**
As title, I'm wondering is there any option or handler before periodic task enqueue?

**Describe the solution you'd like**
Maybe, we can solve this case just like [error-handling](https://github.com/hibiken/asynq/wiki/Periodic-Tasks#error-handling)

**Additional context**
IMO, we can implement this function in Enqueue part, not only for periodic task.

hibiken:
@ChihSeanHsu thank you for creating a feature request.

What kind of action do you need to perform before enqueueing a task? I want to understand the motivation.

ChihSeanHsu:
@hibiken Thanks for your prompt response.
We want to have a metrics to monitor the periodic enqueue count or some other actions (Currently, I have no another idea).
Although, we can just make this metrics by processing task part, but I just wonder is there any pre-action way not post-action.

ChihSeanHsu:
@hibiken Great.
Thanks for your help!
What will the road map be about this small feature?
Or can I just submit a PR about it?

hibiken:
@ChihSeanHsu Feel free to open a PR if you have bandwidth to work on it. Thanks :)

ChihSeanHsu:
Hi @hibiken 
Here is the [PR](https://github.com/hibiken/asynq/pull/476), and feel free to review it thanks.

hibiken:
Closed via #476. This feature will be included in the next release. Thank you @ChihSeanHsu !

PR #476, opened 2022-05-22
Add preHandler and postHandler for Scheduler

ChihSeanHsu:
As discussion [here](https://github.com/hibiken/asynq/issues/456)

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -26,14 +26,16 @@ type Scheduler struct {
 
 	state *serverState
 
-	logger     *log.Logger
-	client     *Client
-	rdb        *rdb.RDB
-	cron       *cron.Cron
-	location   *time.Location
-	done       chan struct{}
-	wg         sync.WaitGroup
-	errHandler func(task *Task, opts []Option, err error)
+	logger      *log.Logger
+	client      *Client
+	rdb         *rdb.RDB
+	cron        *cron.Cron
+	location    *time.Location
+	done        chan struct{}
+	wg          sync.WaitGroup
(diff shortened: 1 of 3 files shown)


scheduler.go
```diff
@@ -26,14 +26,16 @@ type Scheduler struct {
...
-	wg         sync.WaitGroup
-	errHandler func(task *Task, opts []Option, err error)
+	logger      *log.Logger
+	client      *Client
+	rdb         *rdb.RDB
+	cron        *cron.Cron
+	location    *time.Location
+	done        chan struct{}
+	wg          sync.WaitGroup
+	preHandler  func(task *Task, opts []Option)
```
hibiken, line 36: (nit): rename this to preEnqueueFunc

scheduler.go
```diff
@@ -26,14 +26,16 @@ type Scheduler struct {
...
+	postHandler func(info *TaskInfo, err error)
```
hibiken, line 37: (nit): rename this to postEnqueueFunc

hibiken (commented):
Thank you for creating this PR! Could you also add an entry in CHANGELOG.md under "Unreleased"

scheduler.go
ChihSeanHsu: Modified.

scheduler.go
ChihSeanHsu: Modified.

ChihSeanHsu:
> Thank you for creating this PR! Could you also add an entry in CHANGELOG.md under "Unreleased"

Thanks for your remind. I've added it into CHANGELOG.md.

PR #477, opened 2022-05-24
options: add task option for unlimited retry

AlexCuse:
This adds an `UnlimitedRetry()` option that can be used to extend retries indefinitely on a given task if needed.  This would be nice in our system for a subset of tasks we truly don't want to go to archive - it is a bit more direct to test than other current approaches.  I got a good degree of confidence from the unit tests but verified w the following:

```
func main() {
	opt := asynq.RedisClientOpt{Addr: "localhost:6379", DB: 8}
	srv := asynq.NewServer(
		opt,
		asynq.Config{
			RetryDelayFunc: func(n int, e error, t *asynq.Task) time.Duration {
				return 1 * time.Millisecond
			},
		},
	)

	sampleType := "sample-task"

	go func() {
		time.Sleep(1 * time.Second)
		ct := asynq.NewClient(opt)
		_, _ = ct.Enqueue(asynq.NewTask(sampleType, []byte("task with default retries %v\n")))
		_, _ = ct.Enqueue(asynq.NewTask(sampleType, []byte("task with unlimited retries %v\n")), asynq.UnlimitedRetry())
	}()

	mux := asynq.NewServeMux()
	mux.HandleFunc(sampleType, func(ctx context.Context, task *asynq.Task) error {
		f := string(task.Payload())
		rc, _ := asynq.GetRetryCount(ctx)
		fmt.Printf(f, rc)

		if rc > 30 {
			return asynq.SkipRetry
		}
		return fmt.Errorf("force retry check %d", rc+1)
	})

	if err := srv.Run(mux); err != nil {
		log.Fatalf("could not run server: %v", err)
	}
}
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -88,6 +88,11 @@ func MaxRetry(n int) Option {
 	return retryOption(n)
 }
 
+// UnlimitedRetry returns an option to specify the task will be retried indefinitely.
+func UnlimitedRetry() Option {
+	return retryOption(unlimitedRetry)
+}
+
 func (n retryOption) String() string     { return fmt.Sprintf("MaxRetry(%d)", int(n)) }
 func (n retryOption) Type() OptionType   { return MaxRetryOpt }
 func (n retryOption) Value() interface{} { return int(n) }
@@ -297,6 +302,8 @@ const (
 
 	// Default timeout used if both timeout and deadline are not specified.
 	defaultTimeout = 30 * time.Minute
+
+	unlimitedRetry = -1
 )
 
 // Value zero indicates no timeout and no deadline.
--- a/processor.go
+++ b/processor.go
@@ -330,7 +330,7 @@ func (p *processor) handleFailedMessage(ctx context.Context, l *base.Lease, msg 
 		p.retry(l, msg, err, false /*isFailure*/)
 		return
 	}
-	if msg.Retried >= msg.Retry || errors.Is(err, SkipRetry) {
+	if (msg.Retried >= msg.Retry && msg.Retry != unlimitedRetry) || errors.Is(err, SkipRetry) {
 		p.logger.Warnf("Retry exhausted for task id=%s", msg.ID)
 		p.archive(l, msg, err)
 	} else {
--- a/processor_test.go
+++ b/processor_test.go
@@ -291,6 +291,11 @@ func TestProcessorRetry(t *testing.T) {
 	m3 := h.NewTaskMessage("reindex", nil)
 	m4 := h.NewTaskMessage("sync", nil)
 
+	// max retry message
+	m5 := h.NewTaskMessage("", nil)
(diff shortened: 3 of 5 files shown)


khasanovbi:
I need the same functionality. 
Also maybe it would be easier just to use -1 for unlimited retries, I understand that this break the API, but I don't think a lot of people uses negative numbers to disable retries.

hibiken:
@AlexCuse thank you for creating an PR. 

I think we should avoid adding another `Option` that could collide with another Option `MaxRetry`. 
I agree with @khasanovbi that we could treat `MaxRetry(-1)` as unlimited retry. 
The change would require updates to the tools (CLI, Web UI) as well as the core library.

AlexCuse:
Cool good idea I think that should work.  Would the change to the cli / web UI just be to indicate unlimited when we have a negative number @hibiken ?  I think I can get this updated next week.

AlexCuse:
Incidentally I used max int64 in the meantime and noticed it overflowed to -1 in our telemetry stuff.  It was a sign.

yousifh:
I hit the `math.MaxInt64` bug when I wanted to enable "unlimited" retries because the field gets cast to int32 when encoding the message before storing it in Redis. So instead I ended up using `math.MaxInt32`

AlexCuse:
Aha saved me a step :)

I wonder if it makes sense to base the option off int32 as part of this change.

AlexCuse:
example output from asynq task list w/ unlimited retries (-1) on a task:

```
➜  asynq git:(master) ✗ asynq task list --state=retry --queue=default --db 8
ID                                    Type         Payload              Next Retry  Last Error            Last Failed                   Retried  Max Retry  
--                                    ----         -------              ----------  ----------            -----------                   -------  ---------  
4b0fb28b-7f0d-4a5e-bbaa-44a3862dad27  sample-task  non-printable bytes  right now   force retry check 13  Fri Jul  1 07:17:11 EDT 2022  13       unlimited  
9159f9bc-dc75-4d02-9f0e-67332ae1c445  sample-task  non-printable bytes  right now   force retry check 13  Fri Jul  1 07:17:11 EDT 2022  13       25         
```

AlexCuse:
Sorry @hibiken I have been underwater for a bit.  I think this seems OK going to start looking at any asynqmon changes needed when I get a chance.

khasanovbi:
Hi @hibiken any updates with this PR?

AlexCuse:
I think I might owe an asynqmon PR will try and get it opened next week.

AlexCuse:
These changes have been sitting awhile but they were only places I found reference to max_retry at the time: https://github.com/hibiken/asynqmon/pull/286/files

kamikazechaser:
Closing this in favor of using `MaxRetry(math.MaxInt32)`. This should cover majority of use cases.

Issue #487, opened 2022-06-12
How to implement custom panic recovery


mahdimehrabi:
Hi , this package able to **Automatic recovery of tasks in the event of a worker crash** and that's ok but I want to implement custom panic recovery to store panic and error logs in my sentry account.
are we able to set custom logger for asynq or are we able to implement our custom recovery for asynq or asynq have a custom hook when panic happens or ...

yousifh:
An alternative suggestion in case there is hesitation in adding a recover func handler. 

The code in `perform` currently recovers from the panic and returns a generic error. Could it for example wrap a custom error `ErrTaskPanicked` so it might look something like this
```go
var ErrTaskPanicked = errors.New("task panic")

func (p *processor) perform(ctx context.Context, task *Task) (err error) {
	defer func() {
		if x := recover(); x != nil {
			p.logger.Errorf("recovering from panic. See the stack trace below for details:\n%s", string(debug.Stack()))
			_, file, line, ok := runtime.Caller(1)
			if ok && strings.Contains(file, "runtime/") {
				_, file, line, ok = runtime.Caller(2)
			}

			if ok {
				err = fmt.Errorf("%w [%s:%d]: %v", ErrTaskPanicked, file, line, x) // NEW
			} else {
				err = fmt.Errorf("%w: %v", ErrTaskPanicked, x) // NEW
			}
		}
	}()
	return p.handler.ProcessTask(ctx, task)
}
```

And then apps can just change their own `ErrorHandler` to add an if condition for the new error type
```go
if errors.Is(err, asynq.ErrTaskPanicked) {
    // Handle panic errors
} else {
    // Regular errors
}

Just a suggestion in case we want to explore reusing existing error handler rather than introduce a new one

hibiken:
@mahdimehrabi Thank you for opening an issue.

@linhbkhn95 thank you for creating a PR, however, I'd like to discuss the API change here first.

---

The package provides `ErrorHandler` to let users provide error handling logic. We should leverage this existing API by wrapping the recovered panic in a custom error type (`PanicError`) or a sentinel error value (`ErrPanic`)

linhbkhn95:
I see If we use a wrap error to return then the client/user needs to handle this error like a callback function @hibiken

linhbkhn95:
Thank @hibiken, After brainstorming again I decide to use the ErrorHandler for handling panic errors. 
Please review the PR #491

linhbkhn95:
@hibiken can you review it for me?

PR #491, opened 2022-06-15
feat(*): Support for handling panic error 

linhbkhn95:
### Why did we need it?
- Allow users can handle the panic error if they want
Example:
```
func main() {

	srv := asynq.NewServer(
		asynq.RedisClientOpt{
			Addr: redisAddr,
		},
		asynq.Config{
			// Specify how many concurrent workers to use
			Concurrency: 10,
			// Optionally specify multiple queues with different priority.
			Queues: map[string]int{
				"critical": 6,
				"default":  3,
				"low":      1,
			},
			ErrorHandler: asynq.ErrorHandlerFunc(HandleErrorFunc),
		},
	)

	// mux maps a type to a handler
	mux := asynq.NewServeMux()
	mux.HandleFunc(tasks.TypeEmailDelivery, tasks.HandleEmailDeliveryTask)
	mux.Handle(tasks.TypeImageResize, tasks.NewImageProcessor())
	// ...register other handlers...

	if err := srv.Run(mux); err != nil {
		log.Fatalf("could not run server: %v", err)
	}
}

// The ErrorHandlerFunc type is an adapter to allow the use of  ordinary functions as a ErrorHandler.
// If f is a function with the appropriate signature, ErrorHandlerFunc(f) is a ErrorHandler that calls f.
// HandleError calls fn(ctx, task, err)
func HandleErrorFunc(ctx context.Context, task *asynq.Task, err error) {
	if asynq.IsPanicError(err) {
		// Handle panic error here.
	}
}
```
fixed https://github.com/hibiken/asynq/issues/487


imcvampire:
LGTM! Please review and merge it!

hibiken:
@linhbkhn95 Thank you for opening a PR, and sorry for the delay in response.
I'd like to avoid adding yet another callback in the Config.
I think a better solution is to wrap in a custom error so that users can leverage the ErrorHandler to handle panic.

Please discuss the API in the #487.

linhbkhn95:
@hibiken tks for your suggestion. I updated it.

kamikazechaser:
+1 This would be a useful feature.

kamikazechaser (approved):
LGTM, I have been using this feature for a few weeks now.

Maybe `github.com/test-go/testify` is an unnecessary dependency because asynq already uses `github.com/google/go-cmp` in other tests so there will be some little inconsistency.

kamikazechaser:
@hibiken We can merge this.

linhbkhn95:
@hibiken I don't know whether u have a concern about this PR?
I really wanna use this feature this root repo instead of my repo

kamikazechaser:
Thanks @linhbkhn95! It will be included in the next release tag.

PR #505, opened 2022-06-24
feat: add custom unique key option

22dm:
When Enqueue Unique Task, allow the user to customize the UniqueKey instead of using Task Payload by default, because the payload may contain things like trace IDs from upstream, and in some scenarios these tasks should still be considered the same task despite the differences in these fields.

The reason for not using task ID is that TTL is needed to avoid some workers quitting abnormally, making it impossible to continue the task.

### DIFF
--- a/client.go
+++ b/client.go
@@ -44,6 +44,7 @@ const (
 	TimeoutOpt
 	DeadlineOpt
 	UniqueOpt
+	UniqueKeyOpt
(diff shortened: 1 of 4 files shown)


22dm:
Yes, although the projects I'm working on don't use this package now, I still think it's a useful feature.

I have resolved the conflict.

Issue #329, opened 2021-09-23
[Question] Are there TTL set for tasks? How long does it stay in Redis?


seanyu4296:
Asking because i cannot find this in the wiki and others might find it useful.

Specifically, How long do archived tasks stay in Redis so we can keep track or retry manually?

hibiken:
@seanyu4296 Thank you for the question! I should document this somewhere.

Currently these are defined here as constants:

https://github.com/hibiken/asynq/blob/b3ef9e91a9cec1f9890f3f34653aaf6b90388dbd/internal/rdb/rdb.go#L528-L531

In other words, tasks archived 90days ago (or older) will be deleted. If the archive size reaches 10,000, Asynq will start deleting oldest tasks to keep the size at 10,000 or smaller (even if the oldest tasks are less than 90d old). 

Let me know if you have more questions on this :)

seanyu4296:
Thank you again so much for the swift and detailed response @hibiken 

From the code now, there is no way to override it right? correct me if I'm wrong?

seanyu4296:
1. Another follow up question how often does the worker poll to get new tasks mainly for "scheduled" tasks? I'm looking through the code now and i see there is a process to move from scheduled -> pending in forward.go? What's next after that?

2. Is the execution accurate down to the Milliseconds? 

Let's say I enqueued a task at 00:00:00.000 and scheduled it to process after a minute
`asynq.enqueue(..., asynq.ProcessIn(time.Minute * 1)`

Is there a guarantee that it will execute after 00:01:00.000 ? or is there a possibility for it to run a bit early? because of some polling and zscore comparison?

hibiken:
@seanyu4296 Sorry for the delayed response.

Yes, you're right. Currently, user cannot configure `archiveMaxSize` and `archivedExpirationInDays`. Let me know if you have a need to configure these.

As for question #1, after task goes to `pending` it's FIFO from there so all available worker will look at the pending list and start processing in FIFO fashion. 

To answer question #2, no it's not. It's at granularity of seconds. It'll always be *no earlier* than the time specified by `ProcessIn` or `ProcessAt`. It depends on how backed up the pending list is, and also cadence of forwarder (currently every 5s)

These are all good questions, let me know if you have more!

seanyu4296:
> Yes, you're right. Currently, user cannot configure archiveMaxSize and archivedExpirationInDays. Let me know if you have a need to configure these.

So far, there is no need but it would be nice to have. 

[...]

PR #517, opened 2022-07-14
chore(): set archived task TTL

xuyang2:
See https://github.com/hibiken/asynq/issues/329


ScriptProdigy:
Would love to see this option merged in! @hibiken

PR #528, opened 2022-08-19
To Control more Interval

SaltySooda:


pior:
`JanitorInterval` and `HeartbeatInterval` have been added already.

Do you still need the third setting?

kamikazechaser:
I think the use case is very niche if you need to set a very recoverer interval (Sub minute SLA for failed workers). If there is a good reason, we can revisit it again.

Issue #441, opened 2022-04-28
[FEATURE REQUEST] maxArchiveSize  and archivedExpirationInDays  should be configurable at the server level


ghstahl:
## Configurable MaxArchiveSize and ArchivedExpirationInDays 
There are some tasks I don't care if they eventually fail and would prefer, they just get thrown away.

In my application I spin up multiple asynq servers (hosted in a go routines).  Some I want to have failed tasks archived, other servers I don't care.  It's a house keeping thing to not write stuff to REDIS when nobody cares about them.  

I would like to configure maxArchiveSize  and archivedExpirationInDays  in the [config](https://github.com/hibiken/asynq/blob/5c723f597e01e028f84744bff625e1be803e4ad5/server.go#L94)  

Something like this
```go
type Config struct {
  MaxArchiveSize  int
  ArchivedExpirationInDays int
}
```
When I don't care about the tasks, MaxArchiveSize and ArchivedExpirationInDays would be ZERO.

## Alternately
Via @hibiken 
If you are interested in querying all the archived tasks, and delete them programmatically, you can use Inspector.ListArchivedTasks and Inspector.DeleteTask :)

For this I would probably use the asynq scheduled tasks to wipe out those tasks in archives I don't care about.

ghstahl:
The handlers in the design have no knowledge of the context in which they are used.  For a particular server, returning errors is just fine and it's fine to flow into the archive.    For another server, with the handlers behaving the same way, we want the archive excluded.  

The issue that popped out to me going down the sentinel error path is that normal errors emitted from the handler are fine as long as we are in the Retry window.  For the handler to return the "Don't archive me" error is to know that this is the "last" retry and also the business context of which they were called.

I can see the sentinel error technique being used when handlers are designed to have way more knowledge than I give them and needing to know from asynq that we are at the end.

roccoblues:
Hi @hibiken,

as mentioned [here](https://github.com/hibiken/asynq/issues/329#issuecomment-1195356866) I started making those parameters configurable.

However I quickly realized that it's not so easy because there can be at least 3 external triggers of the archiving:
1. The `asynq.processor` in the `asynq.Server`. This is the easy one, here we have the config and can initialize the `rdb.RDB` with the correct values.
2. The `asynq.Inspector` which can be used in custom code.
3. The web interface also using `asynq.Inspector`.

The problem is that I don't see a way to configure all cases together. And if we don't do that the behavior will be totally confusing.

Any ideas?

[...]

PR #534, opened 2022-09-01
[FEATURE REQUEST] maxArchiveSize and archivedExpirationInDays should be configurable at the server level

ghstahl:
https://github.com/hibiken/asynq/issues/441


internal/rdb/rdb.go
```diff
@@ -869,6 +896,9 @@ return redis.status_reply("OK")`)
 // Archive sends the given task to archive, attaching the error message to the task.
 // It also trims the archive by timestamp and set size.
 func (r *RDB) Archive(ctx context.Context, msg *base.TaskMessage, errMsg string) error {
+	if *(r.config.MaxArchiveSize) <= 0 {
+		return nil
```
bojanz, line 900: Don't we need `return r.DeleteTask(msg.Queue, msg.ID)` here, to ensure that the task is deleted instead of just left in the original queue?

kamikazechaser:
@ghstahl Thanks for the PR. Even though it was opened a long time ago, I'd like to see it merged before we merge #743. Are you in a position to update this PR and fix conflicts (I don't have permissions to push to forks)?

internal/rdb/rdb.go
kamikazechaser: This is fine as the author intended that setting anything < 1 should never delete anything. But, another issue is that if the configs are set < 0, a default value of 1 is used that will trigger an archive which is not intended behavior.

kamikazechaser:
Close this as stale. Another PR should be opened in favor of this.

PR #537, opened 2022-09-09
fix redis sentinel url parse

sillydong:
It is actually using SentinelPassword to do sentinel authentication, see from the code
image
If we set Password from url instead of setting SentinelPassword, we will meet such error
image


hibiken:
LGTM. I'll merge once the build passes

PR #543, opened 2022-09-20
add chinese docs

GargantuaX:


kamikazechaser:
I think this is fine. There is a large Chinese community using this library. @GargantuaX are you willing to update the latest changes on the readme?

Issue #549, opened 2022-09-25
Improve performance of enqueueing tasks


yousifh:
When load testing the library at high load, there is opportunity to optimize the enqueue code path. It's currently doing two things

- Add queue name to the set "asynq:queues"
- Enqueue the task

The first call is only necessary the first time, after that the extra network is just wasting resources. We can optimize that by caching which queues we enqueued to and only run the first call whenever the cache expires. On aggregate this will reduce the network calls, Redis commands, and CPU usage of the service and Redis itself. 

On a load test of 100k tasks/min and 16 producer instances

|                                   | Before        | After     |
| ---                             |    ----         |          --- |
| Producer CPU          | 195 mcores             | 173 mcores   |
| Redis CPU                | 290 ms            | 222 ms      |
| Redis cmds/sec       | 8300            |  6600      |

The PR will highlight the change but it's essentially maintaining a map of queue names to their expiration and refreshing each item every 10 secs. On aggregate this heavily reduces the number of network calls needed as shown above. 

The cache has to expire entries because if a queue is deleted, it will be removed from "asynq:queues" set so next time we enqueue to that queue, we will need to add the queue name there.

PR #550, opened 2022-09-25
Improve performance of enqueueing tasks

yousifh:
Shows a potential solution to #549 

Create a map of queue names -> expiration dates in the RDB client. Every time we enqueue a task to queue X, cache queue X in the map and expire it after 10 secs. A mutex is used to protect the access, but there isn't a concern of lock contention as the clients will spend most of their time issuing network calls to enqueue tasks, and acquiring locks to check expiration is very fast. 

The issue highlights the performance improvements but will re-post them here as well
|                                   | Before        | After     |
| ---                             |    ----         |          --- |
| Producer CPU          | 195 mcores             | 173 mcores   |
| Redis CPU                | 290 ms            | 222 ms      |
| Redis cmds/sec       | 8300            |  6600      |

All deployments were done in a normal k8s environment and metrics were measured through Datadog

Caveat: 
This approach keeps the in-memory map of queues forever. It doesn't delete the entry when it expires. This is much simpler to implement, but if there is an app that enqueues tasks to millions of randomly named queues and the app never gets restarted, there is a potential memory leak. However that seems to be an unlikely use case of this library.

### DIFF
--- a/internal/rdb/rdb.go
+++ b/internal/rdb/rdb.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"sync"
 	"time"
 
 	"github.com/go-redis/redis/v8"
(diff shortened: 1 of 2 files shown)


internal/rdb/rdb.go
```diff
@@ -134,6 +145,23 @@ func (r *RDB) Enqueue(ctx context.Context, msg *base.TaskMessage) error {
 	return nil
 }
 
+func (r *RDB) runIfNeeded(queue string, fn func() error) error {
```
yousifh, line 148: This approach uses a simple mutex instead of RW mutex. Contention shouldn't be an issue since most of the time will be spent issuing Redis network calls to enqueue the task payload. This approach also ensures only a single network call is sent to add the queue to "asynq:queues" set for better optimization.

pior:
We are running this in production, and have confirmed the figures posted in the description.
@kamikazechaser I can rebase / rework the implementation if you think that's needed.

kamikazechaser:
@pior Go for it. Might be a good to explore the use of `sync.Map` in place of protecting the map with a mutex.

kamikazechaser:
Reworked in #946.

PR #562, opened 2022-10-20
Add StateChanged to watch state updates

mindon:
implement StateChanged(func(map[string]interface{}), more ...string) for both client and server, to watch state updates with more customized detail, e.g. "completed:result" as default,
e.g. pending:next|task|result


mindon:
there's a better way

PR #563, opened 2022-10-21
fix: set username in redis uri paring

haoqixu:


kamikazechaser:
Similar:

* https://github.com/hibiken/asynq/pull/924

Also check with go-redis

PR #564, opened 2022-10-22
expose semaphore lock count

ajatprabha:
LockCount method return the current cardinality of the set. This is helpful for metric collection purpose.


ajatprabha:
@hibiken, need a review on this.

PR #567, opened 2022-10-28
Sparrow CI pipeline for Alpine Linux (with redis)

melezhik:
Hi devs! I am SparrowCI maintainer. This is simple and quite fast (120 seconds per build) build pipeline for Alpine Linux with redis. If you are interested in Alpine integration, this could be useful. 

An example of successful build - https://ci.sparrowhub.io/report/1453

### DIFF
--- a/sparrow.yaml
+++ b/sparrow.yaml
@@ -1,0 +1,40 @@
+tasks:
+  -
+    name: test
+    default: true
+    language: Bash
+    code: |
+      set -e
+      go version
+      cd source/tools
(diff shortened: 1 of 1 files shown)


pior:
I think we're happy with Github Actions.
cc @kamikazechaser

kamikazechaser:
Yeah, as @pior says, we are fine with GH Actions (Even though it is relatively slower). We are not a very busy repo (couple of commits/week). Thanks anyways.

PR #577, opened 2022-11-23
Add interface for Inspector methods

vladfr:
This is a code structure change, not a functional one. There is a `base.Broker` interface that RDB implements, so I wanted to do the same for the Inspector. 

This PR adds the interface `base.QueueInspector` and moves some types to the `base` package in order to avoid circular dependencies. As expected, RDB already implements both `base.QueueInspector` and `base.Broker`.

**Why?**. To make the CLI and UI work with different broker implementations. I would like to add another broker to this library, backed by MongoDB. `base.Broker` works fine, I have implemented basic features, but it needs this new interface to add support in the CLI and UI.

All tests are passing.

### DIFF
--- a/inspector.go
+++ b/inspector.go
@@ -19,7 +19,7 @@ import (
 // Inspector is a client interface to inspect and mutate the state of
 // queues and tasks.
 type Inspector struct {
-	rdb *rdb.RDB
+	rdb base.QueueInspector
 }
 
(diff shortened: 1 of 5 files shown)


kamikazechaser:
I'm closing this for now (See #667).

Note: There are other libraries out there that allow you to bring your own broker for lesser features as compared to asynq.

Issue #445, opened 2022-04-30
[FEATURE REQUEST] expose backend Health in client & server


a5r0n:
**Is your feature request related to a problem? Please describe.**
to manage app lifecycle, we need to know when to restart some services

**Describe the solution you'd like**
exposing backend services health (like the Redis connection) should do the trick

@hibiken do u accept PR for this?
Any thoughts how to implement?
NB, i quite new in go, so i need some guidance

hibiken:
@a5r0n Thank you for opening this issue!

I want to do a bit more research around how we should do health checking / export health status, etc. 
I'll get back to you soon 👍

gbolo:
> @a5r0n Thank you for opening this issue!
> 
> I want to do a bit more research around how we should do health checking / export health status, etc. I'll get back to you soon +1

@hibiken how about just a basic redis ping/connection check like the server has right now, exposed through the callback function passed in the server config

```
// Config specifies the server's background-task processing behavior.
type Config struct {
...
	// HealthCheckFunc is called periodically with any errors encountered during ping to the
	// connected redis server.
	HealthCheckFunc func(error)

	// HealthCheckInterval specifies the interval between healthchecks.
	//
	// If unset or zero, the interval is set to 15 seconds.
	HealthCheckInterval time.Duration
...
}
```

This seems to work very good for me on the server side. Would love to have the same on the client side ;)

al-bglhk:
> This seems to work very good for me on the server side. Would love to have the same on the client side ;)

How did you do that? Any sample codes please?

pbarnum:
I personally would love to ping Redis (through the Asynq Client/Server) directly instead of relying on Asynq's callback function to do this for me. I have multiple dependencies performing health checks for my service and each one of them provide a direct function call that I can fanout. However for Asynq, I need to rely on the "latest" status returned, which may end up being stale at that point.

pbarnum:
Hey @hibiken, I'm bumping this thread to see if you had a chance to consider my last message. I also tagged this issue in a PR I submitted a few weeks back.

PR #585, opened 2022-12-14
[RFC] Adds Ping() to client/scheduler/server

pbarnum:


pbarnum:
@hibiken, this should close #445

mdaliyan:
Right now I'm pushing a task with an empty payload and then I'm ignoring it when I consume it for the sake of health check. Exposing Ping makes it much easier and cleaner health check than my approach. Please consider increasing the priority for this PR if possible. Thank you.

rapita (approved):

bojanz:
@hibiken What can we do to push this along? I need a client.Ping() to be able to integrate into my app's health check metrics.

arpitflex:
We are deploying an asynq queue alongside our services in a Kubernetes cluster, and a readiness check like this would be very helpful to ensure better Pod restarts by the orchestrator. Are we looking to include this with the 0.25 release @hibiken?

scheduler.go
```diff
@@ -320,3 +320,8 @@ func (s *Scheduler) clearHistory() {
 		}
 	}
 }
+
+// Ping performs a ping against the redis connection.
+func (s *Scheduler) Ping() error {
```
kamikazechaser, line 325: Check for s.srvStateClosed here as well

kamikazechaser (changes requested):

PR #623, opened 2023-02-28
SchedulerEntryID option for registered scheduled job

ilkerkorkut:
Managing autogenerated scheduler job entry ids are very difficult in some cases and it may need costly processes.

Example Scenario: When we try to generate a scheduler job per user basis and wa may want to follow that scheduler job process and unregister on a certain case. So we need to save scheduler job id to on db(need persistency) or cache after that we should check the which user has which scheduler job entry id. So It goes a costly approach. I tried to add SchedulerEntryID Option for Register method.

### DIFF
--- a/client.go
+++ b/client.go
@@ -49,6 +49,7 @@ const (
 	TaskIDOpt
 	RetentionOpt
 	GroupOpt
+	SchedulerEntryIDOpt
 )
 
 // Option specifies the task processing behavior.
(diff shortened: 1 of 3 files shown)


goliatone:
Any feedback or updates on the status of this PR? This would be helpful to provide idempotent registration of cron tasks

ilkerkorkut:
@hibiken do you have any comments on this pr?

Issue #635, opened 2023-03-31
[BUG] Use a custom redis client that will panic if redis close


shuqingzai:
**Describe the bug**
If you implement the `asynq.RedisConnOpt` interface and provide a custom redis client, `Ctrl + C` stops the `asynq.Server` and `asynq.Subscriber` panics

**To Reproduce**
Steps to reproduce the behavior (Code snippets if applicable):
1. Implement the `asynq.RedisConnOpt` interface and provide a custom redis client
2. Instantiate `asynq.Server` and set up a custom redis client
3. Start the Server
4. Stop custom redis client and `asynq.RedisConnOpt` return nil
5. Ctrl + C stops the Server,  `asynq.Subscriber` panics

**Expected behavior**

**Can support custom providing redis client instead of panic**

Because asynq will start multiple redis clients, and redis will also be used in my project, I hope that only one redis client can be instantiated for unified management and reuse clients

**Screenshots**

1. my redis client
> Implement the asynq.RedisConnOpt interface
>

![image](https://user-images.githubusercontent.com/44406033/229015867-ebd7ca78-4f14-4e25-a636-96723f80edef.png)

stop custom redis client  will return nil

3. `asynq.Subscriber` panics
![image](https://user-images.githubusercontent.com/44406033/229016233-9ebc1786-9358-482d-a861-f1551ee53c82.png)

![image](https://user-images.githubusercontent.com/44406033/229017169-cfcc376e-b364-46cf-a530-98207472156b.png)

Because `my redis client` returns nil ??

**Environment (please complete the following information):**
 - OS: [e.g. MacOS, Linux]
 ```shell
MacOS
```
 - Version of `asynq` package [e.g. v1.0.0]
 ```shell
v0.24.0
```

**Additional context**

bfwbbrj:
@shuqingzai My guess would be the [msg](https://github.com/hibiken/asynq/blob/cc777ebdaa62b69bd6e985fa97117b854e7d1cd6/subscriber.go#L83) is nil. And this is related to to the pubsub returned by your custom redis's [Subscribe](https://github.com/hibiken/asynq/blob/cc777ebdaa62b69bd6e985fa97117b854e7d1cd6/internal/rdb/rdb.go#L1451). Since I have no idea which lib you are using, you probably need to check your lib's implementation.

shuqingzai:
[...]

PR #638, opened 2023-04-06
fix: fix panic when redis client closed before server closed

bfwbbrj:
As mentioned in #635 , closing the redis client before shut down asynq will cause panic in [subscriber](https://github.com/hibiken/asynq/blob/cc777ebdaa62b69bd6e985fa97117b854e7d1cd6/subscriber.go#L83)

### DIFF
--- a/subscriber.go
+++ b/subscriber.go
@@ -46,7 +46,7 @@ func newSubscriber(params subscriberParams) *subscriber {
 func (s *subscriber) shutdown() {
 	s.logger.Debug("Subscriber shutting down...")
 	// Signal the subscriber goroutine to stop.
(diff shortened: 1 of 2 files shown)


bfwbbrj:
@hibiken

davidsamblas (approved):

kamikazechaser:
Related to merged change #742.

kamikazechaser:
If the new API (NewClientFromRedisClient) is used, then attempting to close the client through asynq will return an error https://github.com/hibiken/asynq/blob/master/client.go#L321.

A warning has also been added to guide users on this change. I would recommend moving to the new API and handling the closing on your own.

For now, I will close this PR.

Issue #447, opened 2022-05-04
[FEATURE REQUEST] Expose an option to change standard cron parser for a scheduler


SergeyNSW:
**Is your feature request related to a problem? Please describe.**
I'd like to have an ability to run a job every N seconds, the cron package used behind the scene allows to have extended (with seconds) or custom schedule parser, but the async uses default one.
**Describe the solution you'd like**
Is it possible to extend the asynq.SchedulerOpts so that it would accept custom cron.ScheduleParser and pass it as an option to the cron instance?
Thanks

hibiken:
Upon looking, it seems like spec like `"@every 30s"` works without a custom schedule parser. Does that work for you?

SergeyNSW:
yes, it works for me, thanks

acaloiaro:
Hi @hibiken -- I know this issue is closed, but I think it might be worth considering a different option, which some users might find less surprising:

In `scheduler.go`

```go
- cron: cron.New(cron.WithLocation(loc)),
+ cron: cron.New(cron.WithLocation(loc), cron.WithSeconds()),
``` 

Which allows users to supply their cronspecs with the full six fields `* * * * * *`

PR #644, opened 2023-04-12
Support cronspec seconds field

acaloiaro:
This is related to #447 

As a user it can be surprising to discover how to specify asynq tasks with second-level resolution.

This extends the cronspec parser to support the seconds field. Currently, cronspecs must be specified with the `@every Xs` format when they're to be run more frequently than every minute. 

This PR permits the 6th cronspec field (seconds) to be used when specifying periodic asynq tasks. 

E.g.

`@every 1s` -> `* * * * * *`
`@every 30s` -> `*/30 * * * * *`


acaloiaro:
Stale

acaloiaro:
Stale.

PR #657, opened 2023-05-10
Include registration error in the log

khash:
If registration fails, the error is not logged which includes useful information


kamikazechaser:
@khash Are you able to update this PR?

khash:
@kamikazechaser done

kamikazechaser:
@khash See my comment on the review.

khash:
> @khash See my comment on the review.

I'm sorry but I don't see any PR reviews or file comment on this.

kamikazechaser:
Undo this removal:

`s.logger.Debugf("Writing entries %v", entries)`

khash:
> Undo this removal:
> 
> `s.logger.Debugf("Writing entries %v", entries)`

As per the commit message, this log fires every 5 seconds and fills up a lot of the logs with the same message. what value does this add at this frequency?

kamikazechaser:
It is for whoever wants to run the library with **debug** mode. We generally don't remove anything from this library unless there is a very good reason to do so.

khash:
> It is for whoever wants to run the library with **debug** mode. We generally don't remove anything from this library unless there is a very good reason to do so.

While I agree that this is a debug message, I also know that this library is not the only thing that runs in my application. As such, logging the entire payload of a job, every 5 seconds, when there are many other application logs, make debugging very difficult exactly at the time you'd want to look at the flow of a payload of a job from the framework into the job itself. 

I always use **trace** for that purpose not debug, but if the used library in asynq does't support a lower level of logging than debug, I would suggest logging only once of every few times.

If you feel strongly about this issue, feel free to dismiss the PR, I'll continue using the patched version.

PR #667, opened 2023-06-07
Adds Mongo broker

vladfr:
Adds support for Mongo as a broker with support for enqueue, dequeue, schedules, retries and retention; task cleanup via ttl indexes.

This is a draft to see if there's any interest. It's also waiting for #577.


kamikazechaser:
> PS: Current codebase doesn't work.

Yeah this is about it. Even using certain redis forks/compat layers may not work with this library.

PR #673, opened 2023-06-25
fix: queues map init with size

ghosx:
Initializing the queue can improve performance and reduce the number of map expansions. When there are 10 queues, the following is my performance test:

```shell
goos: linux
goarch: amd64
pkg: go-all-in-one/map
cpu: AMD EPYC 7K62 48-Core Processor
BenchmarkWithSize-8 2684252 447.0 ns/op 576 B/op 1 allocs/op
BenchmarkWithoutSize-8 2147931 557.8 ns/op 292 B/op 1 allocs/op
```

### DIFF
--- a/internal/base/base.go
+++ b/internal/base/base.go
@@ -375,7 +375,7 @@ func EncodeServerInfo(info *ServerInfo) ([]byte, error) {
 	if info == nil {
 		return nil, fmt.Errorf("cannot encode nil server info")
 	}
-	queues := make(map[string]int32)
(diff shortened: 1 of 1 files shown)


kamikazechaser:
Not really a fix. But **very minimal** runtime costs improvement.

pior:
Very minimal optimisation, that would be visible on micro-benchmark only.
But since the code is correct, and this is very standard for a map copy, I think we should accept it.
cc @kamikazechaser

Issue #671, opened 2023-06-20
[BUG] tasks before 8 AM are being counted towards the previous day


xmwilldo:
**Describe the bug**
My scheduler program runs in the Asia/Shanghai time zone, and the asynq server also runs in the Asia/Shanghai environment. However, when I start it, I can see that the timestamps in the logs are not in the Asia/Shanghai time zone. I'm not sure if this is causing a problem: the scheduled tasks before 8 AM are being counted towards the previous day, which is not what I want to see.

**To Reproduce**
Steps to reproduce the behavior (Code snippets if applicable):
1. To start the scheduler in the Asia/Shanghai time zone
2. Start  the server
3. Run a scheduled task before 8 AM
4. View the statistical results on the web UI.

**Expected behavior**
The task statistics after 8 AM are for the current day.

**Screenshots**

**Environment (please complete the following information):**
 - OS: Linux
 - Version of `asynq` package v0.22.1

**Additional context**

ghosx:
Do you have reproducible code and screenshots?

naisev:
I have same problem.

hibiken:
@xmwilldo thanks for reporting an issue.

Did you specify your timezone via [`SchedulerOpts.Location`](https://pkg.go.dev/github.com/hibiken/asynq#SchedulerOpts.Location)?

If so, would you mind providing some code snippets to help me investigate further?

xmwilldo:
@hibiken Yes, I tried but it didn't work.
``` go
loc, _ := time.LoadLocation("Asia/Shanghai")
	mgr, err := asynq.NewPeriodicTaskManager(
		asynq.PeriodicTaskManagerOpts{
			PeriodicTaskConfigProvider: &TrainPeriodicTaskProvider{
				ctx:    l.ctx,
				svcCtx: l.svcCtx,
			},
			RedisConnOpt: asynq.RedisClientOpt{
				Addr:     l.svcCtx.Config.Redis.Host,
				Password: l.svcCtx.Config.Redis.Pass,
				DB:       10,
			},
			SchedulerOpts: &asynq.SchedulerOpts{
				Location: loc,
			},
			SyncInterval: 10 * time.Second,
		},
	)
	if err != nil {
		log.Fatalf("could new PeriodicTaskManager: %v", err)
	}
```

xmwilldo:
[...]

PR #683, opened 2023-07-08
Change UTC timezone to system time

ioannidesalex:
Issue #671

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -355,7 +355,7 @@ func TestClientEnqueue(t *testing.T) {
 			desc: "With deadline option",
 			task: task,
 			opts: []Option{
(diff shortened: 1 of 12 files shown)


hibiken:
@ioannidesalex Thanks for the PR. I think we don't want to change the behavior of the server for everyone. I think we should support custom timezone via `Config`.

ioannidesalex:
Personally I always set all my servers to UTC, even though I live at a different timezone. But in my humble opinion, it makes sense to use whatever timezone the server runs at by default.

muhfaris:
I agree with @hibiken 
what if i want to use utc, do i have to change the time in the operating system? @ioannidesalex

ioannidesalex:
Well noted.

huiwanggo:
is there any progress ？

xmwilldo:
is there any progress ？

Issue #686, opened 2023-07-10
[BUG] Latest version of asynq breaks asynq CMD


abezzub:
**Describe the bug**
This is a follow up to https://github.com/hibiken/asynqmon/issues/302 I cannot use `cmd` with asynq:
```
github.com/hibiken/asynq/tools/asynq/cmd
../../pkg/mod/github.com/hibiken/asynq/tools@v0.0.0-20230708040154-123d560a4488/asynq/cmd/root.go:369:20: 
cannot use c (variable of type "github.com/go-redis/redis/v8".UniversalClient) as "github.com/redis/go-redis/v9".UniversalClient value in argument to rdb.NewRDB: 
"github.com/go-redis/redis/v8".UniversalClient does not implement "github.com/redis/go-redis/v9".UniversalClient (missing method ACLDryRun)
```

**To Reproduce**

After upgrading to Asynq 0.24.1 and Asynqmon 0.7.2 I am getting the error above when I am trying to build the app that uses https://github.com/hibiken/asynq/tree/master/tools/asynq. Looks like Redis v8 is still used here:
https://github.com/hibiken/asynq/blob/master/tools/asynq/cmd/root.go#L19

**Environment (please complete the following information):**
 - OS: MacOS
 - Version of `asynq` package: 0.24.1

hibiken:
@abezzub Thanks for reporting. PR welcome for this fix!

PR #687, opened 2023-07-17
Upgrade tools to use redis v9

abezzub:
Upgrade `tools` to use Redis v9. This should resolve https://github.com/hibiken/asynq/issues/686


tools/asynq/cmd/root.go
```diff
@@ -404,18 +404,22 @@ func getTLSConfig() *tls.Config {
...
-//         fmt.Fprintf(w, tmpl, u.Name, u.Addr, u.Age)
-//     }
-// }
+//
+//	printRows := func(w io.Writer, tmpl string) {
+//	    for _, u := range data {
+//	        fmt.Fprintf(w, tmpl, u.Name, u.Addr, u.Age)
+//	    }
+//	}
+//
```
hibiken, line 422: Could you revert this change, I don't think it's related.

hibiken (changes requested):
Thanks for opening a PR. Please address the open comment, then I'll approve this PR.

tools/asynq/cmd/root.go
abezzub: Sure, I assumed it is default go tool formatting

abezzub:
@hibiken I updated the PR

tools/asynq/cmd/root.go
hibiken: we can address this in a separate PR :)

Issue #274, opened 2021-05-20
[BUG] WARN: recoverer: could not move task to archive: NOT FOUND


TopScrew:
**Describe the bug**
WARN: recoverer: could not move task to archive: NOT FOUND

**Expected behavior**
![image](https://user-images.githubusercontent.com/35072495/118931458-a1b35d00-b979-11eb-8ca1-b8294be8162c.png)

**Screenshots**

**Environment (please complete the following information):**
 - OS:  Linux
 - Version of `asynq` v0.17.1

hibiken:
@TopScrew Thank you for reporting this issue.

Could you give me more details so that I can reproduce this bug?

hibiken:
Hi @TopScrew, are you still seeing the issue? Or were you able to figure it out?

hibiken:
Closing this for now since no follow up.
Feel free to reopen the issue if anyone is encountering the same bug.

TheDhejavu:
same issue here: This bug is hard to re-create, i could only think of two things
- When the redis configuration was changed to delete LRU 
- When redis storage was full

<img width="1123" alt="Screen Shot 2023-03-22 at 11 52 00 AM" src="https://user-images.githubusercontent.com/19678915/226881991-0afa699d-b1a0-4150-a978-410793327969.png">

yeqown:
![image](https://github.com/hibiken/asynq/assets/27146233/4815e637-953d-4731-b0df-88db45f939e8)

I got this issue too, I debug and monitor the Redis command, and I got the above screenshot, and refer to the `archiveCmd` lua code:

```go
// KEYS[1] -> asynq:{<qname>}:t:<task_id>
// KEYS[2] -> asynq:{<qname>}:active
// KEYS[3] -> asynq:{<qname>}:lease
// KEYS[4] -> asynq:{<qname>}:archived
// KEYS[5] -> asynq:{<qname>}:processed:<yyyy-mm-dd>
// KEYS[6] -> asynq:{<qname>}:failed:<yyyy-mm-dd>
// KEYS[7] -> asynq:{<qname>}:processed
// KEYS[8] -> asynq:{<qname>}:failed
// -------
// ARGV[1] -> task ID
// ARGV[2] -> updated base.TaskMessage value
// ARGV[3] -> died_at UNIX timestamp
// ARGV[4] -> cutoff timestamp (e.g., 90 days ago)
// ARGV[5] -> max number of tasks in archive (e.g., 100)
[...]

PR #709, opened 2023-07-30
fix(274): listLeaseExpiredCmd ignore tasks those are no longer exist

yeqown:
fix issue #274

### DIFF
--- a/internal/rdb/rdb.go
+++ b/internal/rdb/rdb.go
@@ -1284,7 +1284,10 @@ local res = {}
 local ids = redis.call("ZRANGEBYSCORE", KEYS[1], "-inf", ARGV[1])
 for _, id in ipairs(ids) do
 	local key = ARGV[2] .. id
-	table.insert(res, redis.call("HGET", key, "msg"))
+	local v = redis.call("HGET", key, "msg")
+	if v then
+		table.insert(res, v)
+	end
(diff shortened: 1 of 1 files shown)


yeqown:
For more detail information, I've put them into issue already.

yeqown:
What should I do with this codecov report? @kamikazechaser

kamikazechaser:
> codecov report

Personally, I think it is fine and would ignore it.

Though it would be good to know an easy way to reproduce this bug.

yeqown:
Further more, I found the root cause is in the Lua script `listLeaseExpiredCmd`. while a task key is deleted manually, but it is still in a `asynq:{qname}:lease` key, `listLeaseExpiredCmd` return values would contains a `nil` item, and then be processed which produces the ZERO `TaskMessage`.

yeqown:
@kamikazechaser Can you merge this?

kamikazechaser:
I might need more info on how to reproduce the issue. From #274 it is not very clear on how this is encountered. But the change does look sensible in any case.

kamikazechaser (approved):
LGTM

@hibiken We can merge this

kamikazechaser:
Possibly related to #730.

hibiken (approved):

PR #736, opened 2023-09-07
feat: add publish and getpubsub for streaming

### LINKED ISSUE #437: Added pipelining support to asynq
Added pipelining support to asynq, making least changes on other classes. The idea is capturing all commands sent with pipeline, and replaying them for cluster redirecting.

FogDong:
Fix #437 

This PR adds `publish` and `getPubSub` so that users can use `publish` to stream messages and use `getPubSub` to receive.


kamikazechaser:
Thanks for this. From what I can see this is a feature to track tasks via Redis pub/sub. An alternative to ResultWriter.Write.

However this is not really a fix for #437 and looks very unrelated to it (capturing Redis commands). Could you create a new issue and maybe get feedback on it? Nevertheless, this looks like a good feature to have.

Issue #719, opened 2023-08-07
[Discussion/Proposal] Support last 2 major Go releases only


kamikazechaser:
This is a proposal to only support the last 2 Go releases in asynq. This also aligns with the Go release policy.

If you use Go < 1.19 this could potentially impact you.

## Why?

- Some PR's (especially the refactor ones) rely on new stdlib features
- Some dependency upgrades will not pass CI tests on older Go releases, possibly related to above

## Changes

- Update CI to run tests on 1.19+ or later
- Update go.mod directive to 1.19 or later

##  Refs

- https://go.dev/ref/mod#go-mod-file-go

hibiken:
Thanks for opening this issue with a proposal.
I'm for it -- if others don't have major concerns, we can go ahead and update the README and drop support for older go versions (also make sure to update CI build to drop those go versions)

ghstahl:
I am in a situation where there are no such thing as legacy compilers.   Every docker image that we release gets scanned and we fail soc 2 compliance because of golang.  Which is actually good, it forces use to NEVER get behind.    The week a new golang comes out we upgrade across the board.  

So, my vote is: Never support legacy EVER.  Old stuff had vulnerabilities.

PR #740, opened 2023-09-18
Bound build to the latest two go versions

hibiken:
Discussed in #719


kamikazechaser:
Currently the go.mod version directive is set to 1.14.

Also these are the side effects of updating that to the latest: https://go.dev/blog/toolchain

PR #742, opened 2023-09-19
Implement reusing redis client

jerbob92:
This PR allows you to re-use an existing Redis client with Asynq, this gives you the benifit of less Redis connections and more control over the Redis configuration options since Asynq doesn't provide all the options.

### DIFF
--- a/client.go
+++ b/client.go
@@ -10,11 +10,11 @@ import (
 	"strings"
 	"time"
 
-	"github.com/redis/go-redis/v9"
 	"github.com/google/uuid"
 	"github.com/hibiken/asynq/internal/base"
 	"github.com/hibiken/asynq/internal/errors"
 	"github.com/hibiken/asynq/internal/rdb"
+	"github.com/redis/go-redis/v9"
 )
 
 // A Client is responsible for scheduling tasks.
@@ -25,15 +25,26 @@ import (
 // Clients are safe for concurrent use by multiple goroutines.
 type Client struct {
 	broker base.Broker
+	// When a Client has been created with an existing Redis connection, we do
+	// not want to close it.
+	sharedConnection bool
 }
 
 // NewClient returns a new Client instance given a redis connection option.
 func NewClient(r RedisConnOpt) *Client {
-	c, ok := r.MakeRedisClient().(redis.UniversalClient)
+	redisClient, ok := r.MakeRedisClient().(redis.UniversalClient)
 	if !ok {
 		panic(fmt.Sprintf("asynq: unsupported RedisConnOpt type %T", r))
 	}
-	return &Client{broker: rdb.NewRDB(c)}
+	client := NewClientFromRedisClient(redisClient)
+	client.sharedConnection = false
+	return client
+}
+
+// NewClientFromRedisClient returns a new Client instance given a redis client.
+// Warning: the redis client will not be closed by Asynq, you are responsible for closing.
+func NewClientFromRedisClient(c redis.UniversalClient) *Client {
+	return &Client{broker: rdb.NewRDB(c), sharedConnection: true}
 }
 
 type OptionType int
@@ -150,9 +161,9 @@ func (t deadlineOption) Value() interface{} { return time.Time(t) }
 // TTL duration must be greater than or equal to 1 second.
 //
 // Uniqueness of a task is based on the following properties:
-//     - Task Type
-//     - Task Payload
-//     - Queue Name
+//   - Task Type
+//   - Task Payload
+//   - Queue Name
 func Unique(ttl time.Duration) Option {
 	return uniqueOption(ttl)
 }
@@ -307,6 +318,9 @@ var (
 
 // Close closes the connection with redis.
 func (c *Client) Close() error {
+	if c.sharedConnection {
+		return fmt.Errorf("redis connection is shared so the Client can't be closed through asynq")
+	}
 	return c.broker.Close()
 }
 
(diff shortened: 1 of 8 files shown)


kamikazechaser:
This is already possible without any changes to `asynq`: https://pkg.go.dev/github.com/hibiken/asynq?utm_source=godoc#RedisConnOpt

E.g.

```go
package pool

type (
	SharedRedisPoolOpts struct {
		DSN          string
		MinIdleConns int
	}

	SharedRedisPool struct {
		Client *redis.Client
	}
)

func NewSharedRedisPool(o SharedRedisPoolOpts) (*SharedRedisPool, error) {
	redisOpts, err := redis.ParseURL(o.DSN)
	if err != nil {
		return nil, err
	}

	redisOpts.MinIdleConns = o.MinIdleConns
	redisClient := redis.NewClient(redisOpts)

	return &SharedRedisPool{
		Client: redisClient,
	}, nil
}

// Fulfills asynq's Interface
func (r *SharedRedisPool) MakeRedisClient() interface{} {
	return r.Client
}

// Somewhere else when creating an asynq server asynq.NewServer()
// use *pool.SharedRedisPool
```

jerbob92:
Thanks! I knew it is already possible but it's not really that clear to users how they can do it and it doesn't protect you against accidental closing of the client.

kamikazechaser:
> how they can do it

Right, it is not documented in the wiki. Maybe it should be added there.

> doesn't protect you against accidental closing of the client.

True.

Btw looks like some tests are failing.

jerbob92:
> Btw looks like some tests are failing.
Yep, fixed!

inspector.go
```diff
@@ -28,13 +31,25 @@ func NewInspector(r RedisConnOpt) *Inspector {
 	if !ok {
 		panic(fmt.Sprintf("inspeq: unsupported RedisConnOpt type %T", r))
 	}
+	inspector := NewInspectorFromRedisClient(c)
+	inspector.sharedConnection = false
+	return inspector
+}
+
+// NewFromRedisClient returns a new instance of Inspector.
```
kamikazechaser, line 39: // NewInspectorFromRedisClient returns a new instance of Inspector given a redis.UniversalClient
// Warning: The underlying redis connection pool will not be closed by Asynq, you are responsible for closing it.

client.go
```diff
@@ -25,15 +25,26 @@ import (
...
 		panic(fmt.Sprintf("asynq: unsupported RedisConnOpt type %T", r))
 	}
-	return &Client{broker: rdb.NewRDB(c)}
+	client := NewClientFromRedisClient(redisClient)
+	client.sharedConnection = false
+	return client
+}
+
+// NewClientFromRedisClient returns a new Client instance given a redis client.
+// Warning: the redis client will not be closed by Asynq, you are responsible for closing.
```
kamikazechaser, line 45: // NewClientFromRedisClient returns a new instance of Client given a redis.UniversalClient
// Warning: The underlying redis connection pool will not be closed by Asynq, you are responsible for closing it.

scheduler.go
```diff
@@ -43,15 +43,27 @@ type Scheduler struct {
...
+	redisClient, ok := r.MakeRedisClient().(redis.UniversalClient)
 	if !ok {
 		panic(fmt.Sprintf("asynq: unsupported RedisConnOpt type %T", r))
 	}
+	scheduler := NewSchedulerFromRedisClient(redisClient, opts)
+	scheduler.sharedConnection = false
+	return scheduler
+}
+
+// NewSchedulerFromRedisClient returns a new Scheduler instance given a redis client.
```
kamikazechaser, line 63: // NewSchedulerFromRedisClient returns a new instance of Scheduler given a redis.UniversalClient
// The parameter opts is optional, defaults will be used if opts is set to nil.
// Warning: The underlying redis connection pool will not be closed by Asynq, you are responsible for closing it.

scheduler.go
```diff
@@ -262,7 +274,9 @@ func (s *Scheduler) Shutdown() {
 
 	s.clearHistory()
 	s.client.Close()
```
kamikazechaser, line 276: In the case of `scheduler.Shutdown`, the client is closed by Asynq. If it uses a shared pool, it will return an error (which is unhandled here). Perhaps it should be moved to the following `if` clause?

If the scheduler ends up using a shared redis connection pool, perhaps it should only be stopped. Right now there doesn't seem to be an option to simply Stop/Pause the scheduler though

server.go
```diff
@@ -402,10 +405,19 @@ const (
...
+	redisClient, ok := r.MakeRedisClient().(redis.UniversalClient)
 	if !ok {
 		panic(fmt.Sprintf("asynq: unsupported RedisConnOpt type %T", r))
 	}
+	server := NewServerFromRedisClient(redisClient, cfg)
+	server.sharedConnection = false
+	return server
+}
+
+// NewServerFromRedisClient returns a new Server given a redis client
```
kamikazechaser, line 417: // NewServerFromRedisClient returns a new instance of Server given a redis.UniversalClient
// and server configuration
// Warning: The underlying redis connection pool will not be closed by Asynq, you are responsible for closing it.

kamikazechaser (commented):
Some feedback on godocs and the Scheduler shutdown behavior with this change.

kamikazechaser:
@jerbob92 This is ready to be merged. Could you update the inline docs so that we have a uniform outlook. Thanks.

kamikazechaser:
Merged remote branch into develop 5daa3c52edfad53e3416e784f0c195b1df4bec72 (because I dont have perm to push to forks).

PR #715, opened 2023-08-04
feature: configurable janitor interval and deletion batch size

zhenqianz:

PR #833, opened 2024-03-01
Add JanitorInterval config parameter in server

lsgndln:
**What do you want to do, and how do you expect Asynq to support you with that?**
Hey! I would like to add a new configuration parameter `JanitorInterval` to be able to override the default one of 8 seconds.

**How might this be added to Asynq?**
n/a

**What are possible alternatives?**
n/a

**Are there any disadvantages?**
No, this adds more flexibility.


enzshew (approved):

kamikazechaser:
Duplicate of #715

Issue #774, opened 2023-11-01
[FEATURE REQUEST] Will your add the TraceId option


mason51:
**Is your feature request related to a problem? Please describe.**
I have read the asynq.Option code. There are ten options, but I can not find the trace option.

We use distributed tracing to trace every http api in our system. If I NewTask and Enqueue from a http api, I want to pass the TraceID into the distributed job.

For example:
http://xxx.com/api/buy -> Enqueue TaskA  
                                       -> Enqueue TaskB ->  call http://xxx.com/backend/delivery -> Enqueue TaskD
                                       -> Enqueue TaskC

I want to trace every procedure

shuqingzai:
#547

Issue #547, opened 2022-09-21
[FEATURE REQUEST] Description of the feature request


shuqingzai:
**Is your feature request related to a problem? Please describe.**

Are there plans to support [opentelemetry](https://opentelemetry.io/) in the future ??

**Describe the solution you'd like**

Hope to support opentelemetry.

**Describe alternatives you've considered**

**Additional context**

hibiken:
@shuqingzai Yes, this is definitely something we should consider.

shuqingzai:
@hibiken Hi

Is there any progress on this feature? ? It seems that this project has not released a new version for a long time, and the maintenance is not so active? ?

shuqingzai:
#838

Issue #393, opened 2022-02-01
[FEATURE REQUEST] Add message headers


AlexeyBelezeko:
I'm struggling a bit with passing trace id with the task. Cause there is no way to get it inside middleware without unmarshaling the payload. This adds unnecessary dependencies to middleware code.

I suggest adding `map[string]string` with headers alongside with payload which lets users pass meta info alongside with payload.

P.S.: Big thank you for your work. Sorry if i missed something in the docs and this actually possible.

hibiken:
@AlexeyBelezeko Thank you for opening a feature request!

Is it possible to use [context values](https://pkg.go.dev/context#WithValue)?

AlexeyBelezeko:
Unfortunately not. I inject trace_id and other meta info into context to use it inside Handlers but it didn't save me from exposing the inner structure of concrete handlers to middleware which I'm wanna avoid.

hibiken:
Hmm I see. Maybe similar to this feature request #326?

Could you give me a more concrete example of what you're trying to achieve? Code snippet would be very helpful here :)

AlexeyBelezeko:
In our project we run asynq server and client in two different processes. And to pass traceID i was forced to write middleware like this:
```
type TaskType1 struct {
	TraceID   string
	OtherData []byte
	URL       string
}

type TaskType2 struct {
	TraceID     string
	Name        string
	FullAddress string
	URL         string
}

func TraceInjectorMiddleware(h asynq.Handler) asynq.Handler {
	return asynq.HandlerFunc(func(ctx context.Context, t *asynq.Task) error {
		var traceID string
		switch t.Type() {
		case "type1":
			task := TaskType1{}
			_ = json.Unmarshal(t.Payload(), &task)
			traceID = task.TraceID
		case "type2":
			task := TaskType2{}
			_ = json.Unmarshal(t.Payload(), &task)
			traceID = task.TraceID
		}
		return h.ProcessTask(context.WithValue(ctx, "traceID", traceID), t)
	})
}
```
I suggest adding headers:
```
// Task represents a unit of work to be performed.
type Task struct {
	// typename indicates the type of task to be performed.
	typename string

	// payload holds data needed to perform the task.
	payload []byte

	headers map[string]string

	// opts holds options for the task.
	opts []Option

	// w is the ResultWriter for the task.
	w *ResultWriter
}

func (t *Task) Headers() map[string]string { return t.headers }
```
And this change will help to simplify middleware like this:
```
[...]

Issue #321, opened 2021-08-31
[FEATURE REQUEST] Instrumentation/Tracing for asynq redis client


seanyu4296:
**Is your feature request related to a problem? Please describe.**
- I want to add tracing to the underlying redis client instance using `dd-trace-go`. 
- Is it possible to get access to the underlying redis instance so I can wrap the client with a function like the one below?

```golang
import (
	"github.com/go-redis/redis/v7"
	ddRedis "gopkg.in/DataDog/dd-trace-go.v1/contrib/go-redis/redis.v7"
)

func TraceClient(rClient redis.UniversalClient) {
	ddRedis.WrapClient(
		rClient,
		ddRedis.WithServiceName(config.ServiceName()+"-redis"),
		ddRedis.WithAnalytics(config.Enabled()),
	)
}

```

**Describe alternatives you've considered**
- I am also looking if the way to go about this is creating a `RedisConnOpt` interface that returns a wrapped redis instance with datadog tracing. What do you think?

https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1@v1.33.0/contrib/go-redis/redis.v7#WrapClient

hibiken:
@seanyu4296 thank you for opening this issue!

As I commented in #320 , I'd like to avoid exposing the redis client library (and its version) from `asynq` package API.
I think the alternative you mentioned is a good compromise. You can implement a type that satisfies `RedisConnOpt` interface and provide an object of that type to  `asynq.NewClient` and `asynq.NewServer`. 

Let me know if that works!

seanyu4296:
Got it, will test it out the next coming days and let u know. Are there possible gotchas that you @hibiken  can think atm since I'm not a pro in go?

seanyu4296:
was able to make it work btw @hibiken 😄  see screenshot below

![Screen Shot 2021-09-01 at 10 04 53 PM](https://user-images.githubusercontent.com/11349232/131786966-4b065739-75de-476b-aa10-8beda74ba73d.png)

do you think it will be helpful to add a wiki section on asynq/Tracing

seanyu4296:
also updated this doc to use Stop and Shutdown

https://github.com/hibiken/asynq/wiki/Monitoring-and-Alerting

hibiken:
Also, a heads-up: I just merged PR #298 so you would need to update the redis-client library to v8 when the new release comes out!

seanyu4296:
[...]

PR #838, opened 2024-03-10
feat: add metadata task

alabarjasteh:
A field has been introduced to the `Task` for carrying values across API boundaries. It can be utilized to address cross-cutting concerns, including distributed tracing, as highlighted in issues #547  and #774 .


asynq.go
```diff
@@ -44,19 +48,21 @@ func (t *Task) ResultWriter() *ResultWriter { return t.w }
 
 // NewTask returns a new Task given a type name and payload data.
 // Options can be passed to configure task processing behavior.
-func NewTask(typename string, payload []byte, opts ...Option) *Task {
+func NewTask(typename string, payload []byte, md Metadata, opts ...Option) *Task {
```
kamikazechaser, line 51: This is a breaking change.

kamikazechaser (changes requested):
I would suggest using **functional arguments**. I would refactor this entire PR to use something like:

`asynq.Metadata` which implements `Option`

Anyways the API needs to be discussed first.

kamikazechaser:
Similar PR:

* #606 

Other Issues:

* #393 
* #321

shuqingzai:
#547

michaelelfauz:
any update?

afifurrohman-id:
any update?

PR #865, opened 2024-04-18
feat: add finished handler to server config

### LINKED ISSUE #864: [FEATURE REQUEST] Add callback for task state changes

I'm currently working on a feature that notifies other services when a task has finished.
Unfortunately, I found that asynq doesn't provide an api for this feature.

I currently do this via [middleware](https://github.com/hibiken/asynq/wiki/Handler-Deep-Dive#using-middleware), but it has some flaws and doesn't work well.

One of the key issues is that the life cycle of the task does not actually end when the middleware is executed.
If another service is notified of the end of the task at this point, however, it may still be active when the other service gets the task status from the inspector.

```go
func TaskFinishEventMiddleware() asynq.MiddlewareFunc {
	return func(h asynq.Handler) asynq.Handler {
		return asynq.HandlerFunc(func(ctx context.Context, t *asynq.Task) error {
			var taskId string
			var queueName string
			if val, ok := asynq.GetQueueName(ctx); ok {
				queueName = val
			} else {
				return fmt.Errorf("can not get queue name from task context")
			}
			if val, ok := asynq.GetTaskID(ctx); ok {
				taskId = val
			} else {
				return fmt.Errorf("can not get task id from task context")
			}

			// Send notifications to other services to notify the end of the task
			
			return taskErr
		})
	}
}
```

One solution is to add a global callback function that is called when the task state changes.

```go
var server = asynq.NewServer(redisClientOpt, asynq.Config{
	Queues:   queues,
	Logger:   log.Log,
	LogLevel: logLevel,

	// Add new callback 
	TaskStateChange: func(taskInfo *asynq.TaskInfo) {

	},
})
```

I'm currently using middleware, but it's not working well.

hungtcs:
Hi All!

I've added a FinishedHandler to handle the task after it's finished.
This is a very compromised implementation, as it can fail when posting task status through the Broker, which can cause redis to delay updating the data, at which point the problem I raised in #864 could still occur.

This would probably need to be done by listening for redis state, but I don't know anything about redis, so for now I can only offer this solution.

https://github.com/hibiken/asynq/blob/8df0bfa5834b2e33380ad244925df2fcab1467a9/processor.go#L371-L382

Closes #864

@hibiken Do you have any ideas or plans for this?


kamikazechaser:
@hungtcs Generally this doesn't take into account:

* Failed messages
* Done messages

PR #868, opened 2024-04-22
Add jitter on the processor fetch backoff sleep

pior:
## Changes

This PR adds some jitter to the backoff sleep of the processor fetch loop to avoid workers synchronization, and thus improve worker efficiency and job latency.

## Context

We run dozens of workers, with an execution rate of up to 5K jobs/sec.

We started to suspect that the job workers were significantly inefficient.

1. Changing the TaskCheckInterval from 1s (default) to 100ms has reduced the latency from ~500ms to 40ms on a high traffic worker pool, and 10ms to a low traffic worker pool.

2. Then, adding the jitter has further reduced the latency to 2 ms on both high and low traffic worker pool.

Synchronization on a distributed system is well known, and adding jitter helped avoid it.

Impact of adding a jitter: Queue size:
Screenshot 2024-04-15 at 19 05 32

Latency:
Screenshot 2024-04-15 at 19 05 51

### DIFF
--- a/processor.go
+++ b/processor.go
@@ -181,7 +181,8 @@ func (p *processor) exec() {
 			// Sleep to avoid slamming redis and let scheduler move tasks into queues.
 			// Note: We are not using blocking pop operation and polling queues instead.
 			// This adds significant load to redis.
-			time.Sleep(p.taskCheckInterval)
+			jitter := time.Duration(rand.Intn(int(p.taskCheckInterval)))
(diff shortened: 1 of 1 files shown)


pior:
Rebased on master

kamikazechaser:
@pior Merging this for now. However, we probably need another feature introducing a global rand generator in `processor.go` because it is being used in multiple places. It is probably a good idea to also look into rand/v2.

Related to #916

PR #878, opened 2024-05-01
fix: stop active tasks before server shutdown

kanzihuang:


kamikazechaser:
@kanzihuang The shutdownTimeout will always re-queue jobs that did not complete before it quits. This PR changes behavior in that running tasks are stopped and requeued for next time without allowing the jobs to complete.

```
	// ShutdownTimeout specifies the duration to wait to let workers finish their tasks
	// before forcing them to abort when stopping the server.
	//
	// If unset or zero, default timeout of 8 seconds is used.
```

kanzihuang:
> @kanzihuang The shutdownTimeout will always re-queue jobs that did not complete before it quits. This PR changes behavior in that running tasks are stopped and requeued for next time without allowing the jobs to complete.

After receiving the exit signal, the server does not wait for the task to stop before exiting. The test code is as follows, and the problem can be reproduced:
[test: stop active tasks before server shutdown](https://github.com/hibiken/asynq/pull/878/commits/100bb8987717913a74ca5e997c2c00ba10553417#diff-53170cac6baf0bf02f3aaa3994259ea4cffaaf2498e768fb7e03cf9f752af96cR87)

This PR only addresses the above issues and does not change behavior in that running tasks are stopped and requeued for next time without allowing the jobs to complete.

PR #882, opened 2024-05-07
feature: revoke task

kanzihuang:
Revoke the task to modify task parameters and enqueue new task with the same task id when ProcessTask returns a RevokeTask error.


kamikazechaser:
Thanks @kanzihuang, this is a good feature to have.

P.S I did get your email, but didn't have time to review.

PR #891, opened 2024-06-10
feat: add MinIdleConns to RedisClientOpt struct

sprabowo:
context: i want to have min idle conns config available on asynq. asynq uses go-redis@v9 which support MinIdleConns. 

changelog:
- Add MinIdleConns to RedisClientOpt struct

### DIFF
--- a/asynq.go
+++ b/asynq.go
@@ -284,6 +284,11 @@ type RedisClientOpt struct {
 	// Default is 10 connections per every CPU as reported by runtime.NumCPU.
 	PoolSize int
 
+	// Minimum number of idle connections which is useful when establishing
+	// new connection is slow.
+	// Default is 0. the idle connections are not closed by default.
+	MinIdleConns int
+
 	// TLS Config used to connect to a server.
 	// TLS will be negotiated only if this field is set.
 	TLSConfig *tls.Config
@@ -300,6 +305,7 @@ func (opt RedisClientOpt) MakeRedisClient() interface{} {
 		ReadTimeout:  opt.ReadTimeout,
 		WriteTimeout: opt.WriteTimeout,
 		PoolSize:     opt.PoolSize,
+		MinIdleConns: opt.MinIdleConns,
 		TLSConfig:    opt.TLSConfig,
 	})
 }


sprabowo:
kindly review this small change for exposing MinIdleConns redis config @hibiken @kamikazechaser

kamikazechaser:
@sprabowo You can override the asynq go-redis client and bring your own with `MakeRedisClient` with all custom options that are not otherwise not exposed.

sprabowo:
@kamikazechaser IIRC, should this be handled on app instead of this lib?

i was trying to make my own MustNewClient. it turns out, it does not work.
Screenshot 2024-06-24 at 15 00 32
Screenshot 2024-06-24 at 15 00 01


here are my finding:
1. we can't override client init func
since there's RDB internal pkg call/import, it is not allowed to import `could not import github.com/hibiken/asynq/internal/rdb (invalid use of internal package "github.com/hibiken/asynq/internal/rdb")`. 
2. goredis pkg version will not sync
i use asynq v0.22.1, eventually my goredis package is v7. again i can still use v7 or upgrade but this will not sync 

thus i still propose to just add additional config rather re-implement this CMIIW

sprabowo:
any suggestions @kamikazechaser @hibiken? really appreciate it

linhbkhn95:
Why did you need to make a new client?
type Client struct {
	rdb *rdb.RDB
}
As I understand, You only implement `MakeRedisClient` method as a solution to get individual redis client. After that, you can do like that:
```

type IndividualRedisClientOpt struct {
	Network      string
	Addr         string
	Username     string
	Password     string
	DB           int
	DialTimeout  time.Duration
	ReadTimeout  time.Duration
	WriteTimeout time.Duration
	PoolSize     int
	MinIdleConn  int
	TLSConfig    *tls.Config
}

func (opt IndividualRedisClientOpt) MakeRedisClient() interface{} {
	return redis.NewUniversalClient(&redis.UniversalOptions{
		// ...
		Username:     opt.Username,
		Password:     opt.Password,
		DB:           opt.DB,
		DialTimeout:  opt.DialTimeout,
		WriteTimeout: opt.WriteTimeout,
		PoolSize:     opt.PoolSize,
		MinIdleConns: opt.MinIdleConn,
		TLSConfig:    opt.TLSConfig,
	})
}

func newAsynqClient() *asynq.Client {
	return asynq.NewClient(&IndividualRedisClientOpt{
		// Should fill ...
	})

}
```

pior:
> Perhaps we could document that rather that individually add opts.

This is a reasonable option for sure.
But `MinIdleConn` is a major option, at the same level as PoolSize, I would say.

Other options are probably used less indeed, but PoolSize / MinIdleConn is what are the most used, in my experience.

kamikazechaser:
I would rather document this under performance considerations or how to reuse a go-redis pool with this library in your application than expose a go-redis option. Why?

* The underlying default is also 0, this change will keep it at 0 as well and the only benefit is that we have exposed this option as available in the docs.

@sprabowo You can do something like:

```go
type (
	RedisPoolOpts struct {
		DSN          string
		MinIdleConns int
	}

	RedisPool struct {
		Client *redis.Client
	}
)

func NewRedisPool(o RedisPoolOpts) (*RedisPool, error) {
	redisOpts, err := redis.ParseURL(o.DSN)
	if err != nil {
		return nil, err
	}
	redisOpts.MinIdleConns = o.MinIdleConns

	redisClient := redis.NewClient(redisOpts)

	return &RedisPool{
		Client: redisClient,
	}, nil
}

// Interface adapter for asynq to resuse the same Redis connection pool.
func (r *RedisPool) MakeRedisClient() interface{} {
	return r.Client
}
```

kamikazechaser:
https://github.com/hibiken/asynq/commit/5daa3c52edfad53e3416e784f0c195b1df4bec72

^ In the upcoming release

Issue #510, opened 2022-07-08
scheduler entries ttl maybe too short


xuyang2:
Currently the tick interval of `Scheduler.runHeartbeater()` and the ttl of `WriteSchedulerEntries()` are both `5*time.Second`

scheduler entries is prone to expire when redis is slow to respond

Is it acceptable to increase the ttl of WriteSchedulerEntries() ?

hibiken:
@xuyang2 thank you for reporting this issue!
I agree. I think it makes sense to increase the TTL to 10s to allow for a miss in heartbeat.

tschaub:
I'm seeing behavior that looks like tasks are not getting scheduled.  I'm having trouble putting together a reproducible test case - only seeing the issue in a production environment and not in tests or running locally.  I'm wondering if this TTL issue or some other known issue might be responsible.

Are others seeing cases where tasks are sometimes not getting run or scheduled (no error from enqueue, but the handler is never called with the task)?

lamhieo02:
> @xuyang2 thank you for reporting this issue! I agree. I think it makes sense to increase the TTL to 10s to allow for a miss in heartbeat.

Do you have a new version to handle this?

PR #896, opened 2024-07-02
Fix for issue 510

lamhieo02:
handle for issue 510: https://github.com/hibiken/asynq/issues/510

image

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -10,11 +10,11 @@ import (
 	"sync"
 	"time"
 
-	"github.com/redis/go-redis/v9"
 	"github.com/google/uuid"
 	"github.com/hibiken/asynq/internal/base"
 	"github.com/hibiken/asynq/internal/log"
 	"github.com/hibiken/asynq/internal/rdb"
+	"github.com/redis/go-redis/v9"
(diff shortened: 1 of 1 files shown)


scheduler.go
```diff
@@ -299,7 +299,7 @@ func (s *Scheduler) beat() {
 		entries = append(entries, e)
 	}
 	s.logger.Debugf("Writing entries %v", entries)
-	if err := s.rdb.WriteSchedulerEntries(s.id, entries, 5*time.Second); err != nil {
+	if err := s.rdb.WriteSchedulerEntries(s.id, entries, 10*time.Second); err != nil {
```
linhbkhn95, line 302: should we set it as const instead ?

pior: This is not a regression, it makes sense to me to follow the pattern.

Constants are useful of course, especially when there is an established pattern, otherwise it's mostly an indirection that slightly reduces readability.

pior: @linhbkhn95 My bad, it _is_ already the pattern: 
https://github.com/hibiken/asynq/blob/c97652d408e04eeb944e7ec87cddaae77a5780aa/server.go#L415-L419

PR #914, opened 2024-09-03
Fix memory leaks caused by time.After

### LINKED ISSUE #913: [BUG] Memory leaks caused by time.After
**Describe the bug**
timer.After() creates a new timer object  each time it is called. Which can cause a memory leak in the for loop.
https://github.com/hibiken/asynq/blob/d04888e74845fc7434c6a50755e62ec6ffda5911/subscriber.go#L61-L74
https://github.com/hibiken/asynq/blob/d04888e74845fc7434c6a50755e62ec6ffda5911/syncer.go#L60-L85
**To Reproduce**

**Expected behavior**
Creating the timer object before the for loop.

**Screenshots**

**Environment (please complete the following information):**
 - OS: Linux
 - Version of `asynq` package v0.24.1

**Additional context**

biningo:
Fixes #913

### DIFF
--- a/subscriber.go
+++ b/subscriber.go
@@ -8,9 +8,9 @@ import (
 	"sync"
 	"time"
 
-	"github.com/redis/go-redis/v9"
 	"github.com/hibiken/asynq/internal/base"
 	"github.com/hibiken/asynq/internal/log"
+	"github.com/redis/go-redis/v9"
 )
 
 type subscriber struct {
@@ -58,12 +58,15 @@ func (s *subscriber) start(wg *sync.WaitGroup) {
 			err    error
 		)
 		// Try until successfully connect to Redis.
+		timer := time.NewTimer(s.retryTimeout)
+		defer timer.Stop()
 		for {
 			pubsub, err = s.broker.CancelationPubSub()
 			if err != nil {
 				s.logger.Errorf("cannot subscribe to cancelation channel: %v", err)
 				select {
-				case <-time.After(s.retryTimeout):
+				case <-timer.C:
+					timer.Reset(s.retryTimeout)
 					continue
 				case <-s.done:
 					s.logger.Debug("Subscriber done")
--- a/syncer.go
+++ b/syncer.go
@@ -57,6 +57,8 @@ func (s *syncer) start(wg *sync.WaitGroup) {
 	go func() {
 		defer wg.Done()
 		var requests []*syncRequest
+		timer := time.NewTimer(s.interval)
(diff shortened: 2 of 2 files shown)


subscriber.go
l0nax, line 1: Only starting with Go 1.23 unreferenced and unstopped `time.Ticker` will be garbage collected.

Since the min. Go version of this project is Go 1.20, you should call `defer ticker.Stop()`. Otherwise, this will leak the ticker.

In both cases, the changed code does not behave the same as the original one: `time.Ticker` publishes in a defined Interval $I$. You have to reset the ticker every time you start the next "cycle. "Otherwise, the timeout you're waiting for may be shorter than expected.

l0nax (changes requested):

subscriber.go
biningo: Thank you for your suggestions! : )  
I'll fix it!

**About Go version**
As you said, using time.After in Go 1.23 and later versions does not cause memory leaks, but this project is using Go 1.20.

**About ticker or timer**
The ticker sends values to the channel at a fixed rate. If the task execution time is longer than the ticker interval, the next task will start immediately after the previous task finished.
But the timer can be reset after the task has finished. No matter how long the previous task is running, the next task will start after the interval.

huangc28:
Why haven't we merge this PR?

kamikazechaser:
Thanks @biningo, we will migrate this library to v1.23 soon.

syncer.go
```diff
@@ -57,6 +57,8 @@ func (s *syncer) start(wg *sync.WaitGroup) {
 	go func() {
 		defer wg.Done()
 		var requests []*syncRequest
+		timer := time.NewTimer(s.interval)
```
kamikazechaser, line 60: In the syncer, would it be more appropriate for the initialized Interval to be 0 while the Reset to then use the s.interval value?

kamikazechaser:
If we were to move this library to 1.23, this change wouldn't be required at all. I am not sure if this optimization is worth it (at the expense of code complexity) for what could possibly be 2 months (or even less).

cc/ @pior @l0nax

biningo:
> If we were to move this library to 1.23, this change wouldn't be required at all. I am not sure if this optimization is worth it (at the expense of code complexity) for what could possibly be 2 months (or even less).
> 
> cc/ @pior @l0nax

Upgrading to Go 1.23 is a better solution.

kamikazechaser:
Sounds good. I'll close this for now. Thanks for highlighting this though!

PR #916, opened 2024-09-12
perf: Reuse rand for shuffling queues

l0nax:
Instead of creating a new `rand.Rand` every time we need the next queues, we create it once and reuse it.

The reason for this change comes from monitoring our application: After a few minutes, the `github.com/hibiken/asynq.(*processor).queues` method will consume a tremendous amount of CPU time and create a lot of garbage.

Since the `queue` method is called in its own goroutine, there is no need to add any synchronization primitives.

Here is the profiling flame graph of our application in a time window of 5 minutes:
image

### DIFF
--- a/processor.go
+++ b/processor.go
@@ -70,6 +70,9 @@ type processor struct {
 
 	starting chan<- *workerInfo
 	finished chan<- *base.TaskMessage
+
+	// queueShuffleRand holds the random source to shuffle the queues.
+	queueShuffleRand *rand.Rand
 }
 
 type processorParams struct {
@@ -119,6 +122,7 @@ func newProcessor(params processorParams) *processor {
 		shutdownTimeout:   params.shutdownTimeout,
 		starting:          params.starting,
 		finished:          params.finished,
+		queueShuffleRand:  rand.New(rand.NewSource(time.Now().UnixNano())),
 	}
 }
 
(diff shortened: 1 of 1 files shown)


processor.go
```diff
@@ -119,6 +122,7 @@ func newProcessor(params processorParams) *processor {
 		shutdownTimeout:   params.shutdownTimeout,
 		starting:          params.starting,
 		finished:          params.finished,
+		shuffleRand:       rand.New(rand.NewSource(time.Now().UnixNano())),
```
pior, line 125: `rand.NewSource` is not thread-safe ([code](https://cs.opensource.google/go/go/+/refs/tags/go1.23.2:src/math/rand/rng.go;l=238-252)).

This is fine now since only `queues()` uses it, and `queues()` is only called in a single single-threaded context.

But this seems bug-prone, as it would be tempting to re-use it in a different context.

Maybe naming the field `queueShuffleRand` and adding a comment about thread-safety is enough.

l0nax: I thought the comment I added to the field should be clarifying enough – I renamed it since it makes it clearer.

Thanks!

kamikazechaser:
@l0nax Checkout my comment on #868.

l0nax:
@kamikazechaser If it is okay with the project I can bump the Go version 1.23 and introduce a global rand generator.
I didn't want to do that because I didn't know if it was okay to upgrade the Go version.

kamikazechaser:
We guarantee to support the last 2 versions here: https://github.com/hibiken/asynq?tab=readme-ov-file#quickstart

And since math/rand/v2 was introduced in 1.22...It should be ok.

kamikazechaser:
Added in #944. Thanks @l0nax !

PR #918, opened 2024-09-27
Add task enqueue command to cli

mboorstin:
Add a `task enqueue` command to the CLI to allow manually enqueuing a task with specified payload and various options. This is useful when manually interacting with asynq or attempting to trigger specific actions.

### DIFF
--- a/tools/asynq/cmd/task.go
+++ b/tools/asynq/cmd/task.go
@@ -53,6 +53,24 @@ func init() {
 	taskRunCmd.MarkFlagRequired("queue")
 	taskRunCmd.MarkFlagRequired("id")
 
+	taskCmd.AddCommand(taskEnqueueCmd)
+	taskEnqueueCmd.Flags().StringP("type_name", "t", "", "type name to enqueue the task as (required)")
+	taskEnqueueCmd.Flags().StringP("payload", "l", "", "payload to enqueue (required)")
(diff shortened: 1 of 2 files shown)


tools/asynq/cmd/task.go
```diff
@@ -53,6 +53,24 @@ func init() {
 	taskRunCmd.MarkFlagRequired("queue")
 	taskRunCmd.MarkFlagRequired("id")
 
+	taskCmd.AddCommand(taskEnqueueCmd)
+	taskEnqueueCmd.Flags().StringP("type_name", "t", "", "type name to enqueue the task as (required)")
+	taskEnqueueCmd.Flags().StringP("payload", "l", "", "payload to enqueue (required)")
+	// The following are the various OptionTypes; if not specified we won't pass them so that composeOptions()
+	// can apply its own defaults
```
mboorstin, line 60: Parameter names are chosen to match the names at https://github.com/hibiken/asynq/blob/master/client.go#L67 . The only exception is `id`; the `Option` is named `TaskID` but all of the other commands here use `id` so I figured it was better to stick with that.

mboorstin:
@kamikazechaser @hibiken I've found this added command to be helpful; please let me know if you have any comments on it.  Thank you!

kamikazechaser:
Looks good.

Issue #389, opened 2022-01-28
Is it necessary for the `LREM` argument count = 0 ?


ryan961:
> `LREM`:  Removes the first count occurrences of elements equal to element from the list stored at key
....
> count = 0: Remove all elements equal to element.

In the case of a certain amount of data, it will cause a lot of slow queries, which will cause Redis CPU to explode, and this will enter a vicious cycle that can't be handled.
https://github.com/hibiken/asynq/blob/dff2e3a33639338c648599bf6ea83f835feec4bf/internal/rdb/rdb.go#L312
https://github.com/hibiken/asynq/blob/dff2e3a33639338c648599bf6ea83f835feec4bf/internal/rdb/rdb.go#L344
...

hibiken:
@ryalu Thank you for reporting this issue!

These `LREM` operations are executed against the `asynq:{qname}:active` list which can only grow up to the number specified by `Config.Concurrency` so should be bounded to smallish number (e.g. ~100 or so).

But you made a good point, that this count may only need to be set to 1 since we only need to remove just one element from the list during normal operation. I didn't put too much thoughts there but I wanted to guarantee that we remove the id from the list. We can run benchmark tests with different concurrency settings, and see if changing this to 1 would make it more performant.

PR #921, opened 2024-10-14
Call LREM with count=1 on the active queue item

pior:
When a task is done, marked as completed, archived or requeued, it is removed from the `active` list in Redis.
This operation is done by a `LREM` operation with `count = 0` (remove **all** elements matching the value).

For 100 concurrently active tasks or so, `LREM` is not an issue.
For 1000 concurrent active tasks or more, the `O(N+M)` complexity of LREM seems to become a significant waste of Redis cpu time.

It represents almost 50% of the the Redis CPU time in some of our queues.

Note: we plan to test this change in a realistic environment, which is why the PR is in Draft.

Issue: https://github.com/hibiken/asynq/issues/389

### Solution

Change the LREM count parameter from 0 to -1 (deletes up to 1 element, starting from the tail, [doc](https://redis.io/docs/latest/commands/lrem/)).

Tasks pushed with RPOPLPUSH should generally be at the tail when they are done/completed.

### DIFF
--- a/internal/rdb/rdb.go
+++ b/internal/rdb/rdb.go
@@ -275,7 +275,7 @@ func (r *RDB) Dequeue(qnames ...string) (msg *base.TaskMessage, leaseExpirationT
 // ARGV[2] -> stats expiration timestamp
 // ARGV[3] -> max int64 value
 var doneCmd = redis.NewScript(`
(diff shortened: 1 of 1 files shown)


pior:
So, @hurracrane raised a good point about a possible memory leak in Redis, in case the active list contains duplicate.
This change could uncover a subtle bug, and we don't want to take this risk at the moment.

PR #924, opened 2024-10-25
Need to support redis sentinel username.

Skwol:
Redis sentinel deploy assumes that there should be several users, one for auth in sentinel for getting proper redis instance address and the other one for actual redis auth. Need to support this.

### DIFF
--- a/asynq.go
+++ b/asynq.go
@@ -316,6 +316,9 @@ type RedisFailoverClientOpt struct {
 	// https://redis.io/topics/sentinel.
 	SentinelAddrs []string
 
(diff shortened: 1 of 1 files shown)


Skwol:
Looks like it requires golang update, I see that there is a [PR](https://github.com/hibiken/asynq/pull/925) exactly for update. I guess I can wait for this PR to be merged and rebuild this one after that.

kamikazechaser:
@Skwol Merged the other PR.

Skwol:
I've rebased with the master, so now CI should work fine

PR #925, opened 2024-10-25
Bump to Go 1.22 and 1.23

pior:
## Context

We want to use math/rand/v2, introduced in 1.22.

> We guarantee to support the last 2 versions here: https://github.com/hibiken/asynq?tab=readme-ov-file#quickstart

https://github.com/hibiken/asynq/pull/916#issuecomment-2437054813

## Changes 

- Bump minimal Go version in go.mod to 1.22
- Run CI on 1.22 and 1.23


kamikazechaser:
Ref: https://github.com/hibiken/asynq?tab=readme-ov-file#quickstart

PR #926, opened 2024-10-25
Call all context cancelFunc in processor.go

pior:
## Context

Similar change as https://github.com/hibiken/asynq/pull/145

We are potentially leaking context goroutines in requeue(), markAsComplete(), markAsDone(), retry(), archive() methods.

Those methods are not doing anything weird with the context, they are simply calling `r.runScript(ctx, ...)`, we can safely cancel this context as usual.

## Changes

- Add `defer cancel()`


kamikazechaser (approved):
Agreed. Always good to call cancel on context withDeadline/withTimeout.

kamikazechaser:
Benchstat CI is failing, but that is not necessary for this change.

PR #870, opened 2024-04-24
Remove deprecated protobuf package

aradwann:
removing deprecated github.com/golang/protobuf
and applying go fmt
closes #855

PR #927, opened 2024-10-25
Run golangci-lint in CI

pior:
## Context

We want to run linters in CI

## Changes

- Add a new Github Action workflow to run the official golangci-lint action, with the latest golangci-lint release
- Fix linting errors when possible
- Locally disable the failing linter on the deprecated protobuf API (since we should upgrade)

The fix for the context.CancelFunc errors has its own PR since it's a bug: https://github.com/hibiken/asynq/pull/926


kamikazechaser (approved):
Looks good.

Related to #870

PR #946, opened 2024-10-28
Improve performance of enqueueing tasks

### LINKED ISSUE #549: Improve performance of enqueueing tasks

pior:
## Context

Before enqueueing **every** task, the name of the queue is published to a Redis Set key (`base.AllQueues`).
This is only used by the inspector for inspection/monitoring purposes (not for the processing of tasks).

Two consequences:
- Enqueuing one task requires two round-trip to Redis
- The same Redis set key is updated constantly, causing significant Redis load

The issue #549 has more details about the impact on performance.
We've confirmed the figures in production.

Fixes #549
Rework of #550

## Changes

- Update the RDB client to only write to `base.AllQueues` once per worker

Notes:
- This light implementation does not protect against race-conditions, which is not an issue.
- Writing to `base.AllQueues` will still be retried until it worked.

### DIFF
--- a/internal/rdb/inspect.go
+++ b/internal/rdb/inspect.go
@@ -10,9 +10,9 @@ import (
 	"strings"
 	"time"
 
-	"github.com/redis/go-redis/v9"
 	"github.com/hibiken/asynq/internal/base"
 	"github.com/hibiken/asynq/internal/errors"
+	"github.com/redis/go-redis/v9"
 	"github.com/spf13/cast"
 )
 
@@ -1832,6 +1832,7 @@ func (r *RDB) RemoveQueue(qname string, force bool) error {
 		if err := r.client.SRem(context.Background(), base.AllQueues, qname).Err(); err != nil {
 			return errors.E(op, errors.Unknown, err)
 		}
+		r.queuesPublished.Delete(qname)
 		return nil
 	case -1:
 		return errors.E(op, errors.NotFound, &errors.QueueNotEmptyError{Queue: qname})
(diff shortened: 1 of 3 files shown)


kamikazechaser (commented):
If the queue is removed via RemoveQueue, it will not be added back on the next enqueue because it will still exist in the map. We need to also remove it from the map on any `SREM("asynq::queues", qname)`

kamikazechaser:
> It works if a task is enqueued to the queue in the same process, but won't work for a different process.

That's right. I think the tradeoff is fair.

kamikazechaser (approved):

PR #956, opened 2024-11-04
Add the scheduler option HeartbeatInterval

pior:
## Context

From https://github.com/hibiken/asynq/issues/510:
> Currently the tick interval of Scheduler.runHeartbeater() and the ttl of WriteSchedulerEntries() are both 5*time.Second
> scheduler entries is prone to expire when redis is slow to respond

Redo of https://github.com/hibiken/asynq/pull/896

Fixes https://github.com/hibiken/asynq/issues/510

## Changes

- Add option `SchedulerOpts.HeartbeatInterval`
- Change default scheduler heartbeat interval from 5s to 10s

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -335,7 +335,7 @@ func (s *Scheduler) beat() {
 		entries = append(entries, e)
 	}
 	s.logger.Debugf("Writing entries %v", entries)
-	if err := s.rdb.WriteSchedulerEntries(s.id, entries, s.heartbeatInterval); err != nil {
+	if err := s.rdb.WriteSchedulerEntries(s.id, entries, s.heartbeatInterval*2); err != nil {
 		s.logger.Warnf("Scheduler could not write heartbeat data: %v", err)
 	}
 }
--- a/scheduler.go
+++ b/scheduler.go
@@ -26,16 +26,17 @@ type Scheduler struct {
 
 	state *serverState
 
-	logger          *log.Logger
-	client          *Client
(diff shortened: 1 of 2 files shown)


scheduler.go
```diff
@@ -317,7 +335,7 @@ func (s *Scheduler) beat() {
 		entries = append(entries, e)
 	}
 	s.logger.Debugf("Writing entries %v", entries)
-	if err := s.rdb.WriteSchedulerEntries(s.id, entries, 5*time.Second); err != nil {
+	if err := s.rdb.WriteSchedulerEntries(s.id, entries, s.heartbeatInterval); err != nil {
```
kamikazechaser, line 338: From the original issue:

> scheduler entries are prone to expire when redis is slow to respond

Even though this just affects the inspector, there is a small chance that the SchedulerEntries may be empty because they expired before the next beat ran. I would propose it to be such that redisKey ttl > beat tick.

pior: Absolutely, I did not want to tackle this in the same PR, but I think we should have `ttl = s.heartbeatInterval * 2`.

I don't have strong opinions on this, what option do you prefer:
- `ttl = s.heartbeatInterval * 2`
- `ttl = s.heartbeatInterval + time.Second`
- Or a second option

kamikazechaser: > ttl = s.heartbeatInterval * 2

kamikazechaser (approved):

PR #958, opened 2024-11-06
feat(periodic_task_manager): Add RedisUniversalClient support

daixijun:


daixijun:
@kamikazechaser There is no duplication, and #742 does not implement RedisUniversalClient reuse in PTM

kamikazechaser:
I see, that's right. Could you stick to the same API that is in #742?

kamikazechaser (approved):
LGTM. There is no better way to change the API. Thanks.

PR #976, opened 2024-12-03
feat: allow custom broker

asdfhuang:
This PR decouples `Broker` interface from redis and allows creating `Server` and `Client` directly from a broker instance. Which makes it possible for custom brokers.

For exmaple: Redis from our infrastructure doesn't support pubsub. But I can inherit RDB and implement the pub/sub in another way to make asynq work without a huge amount of work.

### DIFF
--- a/client.go
+++ b/client.go
@@ -44,7 +44,13 @@ func NewClient(r RedisConnOpt) *Client {
 // NewClientFromRedisClient returns a new instance of Client given a redis.UniversalClient
 // Warning: The underlying redis connection pool will not be closed by Asynq, you are responsible for closing it.
 func NewClientFromRedisClient(c redis.UniversalClient) *Client {
-	return &Client{broker: rdb.NewRDB(c), sharedConnection: true}
+	return NewClientFromBroker(rdb.NewRDB(c))
+}
+
+// NewClientFromBroker returns a new instance of Client given a broker.
+// Warning: The underlying broker will not be closed by Asynq, you are responsible for closing it.
+func NewClientFromBroker(b base.Broker) *Client {
+	return &Client{broker: b, sharedConnection: true}
 }
 
 type OptionType int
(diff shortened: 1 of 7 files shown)


asdfhuang:
There are some known issues for this PR:
- Inspector and Scheduler are locked to redis broker now, they might not work if we use custom borkers.
- Using type alias to export Broker is declared from internal package is a temporary solution.

kamikazechaser:
What redis-like broker is this? Is it one of the forks or is it one that implements the Redis wire protocol?

kamikazechaser:
The idea is good. Though we wouldn't land this in master immediately because:

1. We would need to move the types into a publicly importable folder e.g. `pkg`. I like the short term fix, but it makes this library look hacky. We have already introduced some hacky bits which are starting to show (see #977).
2. Work on supporting the scheduler and inspector to use the custom Broker (if possible at all).
3. We have the problem of go-redis which is now transferred to Redis corp. I'm not even sure if there will be guaranteed support for redis forks and clones (https://news.ycombinator.com/item?id=42239607).

asdfhuang:
> What redis-like broker is this? Is it one of the forks or is it one that implements the Redis wire protocol?

It's a redis fork that maintained and operated by our infra team. The problem is that there's always a proxy in front of the redis instances, and the proxy doesn't support pubsub.


> 1. We would need to move the types into a publicly importable folder e.g. `pkg`. I like the short term fix, but it makes this library look hacky. We have already introduced some hacky bits which are starting to show (see [fix: NewScheduler incorrectly creates underlying Client, closing broker properly #977](https://github.com/hibiken/asynq/pull/977)).
> 2. Work on supporting the scheduler and inspector to use the custom Broker (if possible at all).
> 3. We have the problem of go-redis which is now transferred to Redis corp. I'm not even sure if there will be guaranteed support for redis forks and clones (https://news.ycombinator.com/item?id=42239607). (Though I see you changes to `Client` kind of address this, but I'm not sure it addresses everywhere else in the library which ends up using go-redis, repeation  of my 2nd point)

for the 1st& 2nd point, I'll try to upload a new patch to resolve them.

kamikazechaser:
> It's a redis fork

Ah ok. It should generally be fine for now. I have started a discussion around this in #981.

PR #988, opened 2024-12-16
Allow cancel without retry

thedmdim:
Previously we couldn't cancel task without sending it to retry queue. Despite the fact that handler returned `SkipRetry` error, it didn't work correctly. Read from `<-ctx.Done()` in `processor.go:246` caused by canelation were and will be always earlier than returned `SkipRetry` from handler. That's why task always went to retry queue.

With these changes you are able to cancel task so it will be archived instead of going to retry :)

### DIFF
--- a/processor.go
+++ b/processor.go
@@ -336,15 +336,18 @@ func (p *processor) handleFailedMessage(ctx context.Context, l *base.Lease, msg 
 		p.errHandler.HandleError(ctx, NewTask(msg.Type, msg.Payload), err)
 	}
 	switch {
-	case errors.Is(err, RevokeTask):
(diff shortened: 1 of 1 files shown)


kamikazechaser:
The downside of this approach is that every context cancellation will result in the task being cancelled.. There are some cases where you'd want a certain context cancellation to be retry-able. Right now every cancellation will be retried and I believe most users are used to this behaviour by now.

With 1.20, this kind of checking is possible: https://pkg.go.dev/context#WithCancelCause coupled with `errors.Is(context.Cause(ctx), ErrMyRetryableCancelledCtx)`. This will allow a better experience.

I'd like to keep this PR open and have more discussions around it with more people before we agree on the best way to handle this.

PR #1005, opened 2025-01-10
fix compatible with redis server verison 4.0

popstk:
as readme say `Version 4.0 or higher is required`, but  `zrange xxx byscore ` add from version 6.2.0
>> https://redis.io/docs/latest/commands/zrange/
>> Starting with Redis version 6.2.0: Added the REV, BYSCORE, BYLEX and LIMIT options.

---
in addition, 
https://github.com/hibiken/asynq/blob/489e21920b92ae6acfc19c54de91166e56817620/processor.go#L381
``` go
	err := p.broker.Archive(ctx, msg, e.Error())
	if err != nil {
```
The `err` do not print in log,  which can help debug


kamikazechaser:
See #1076. [ZRANGEBY](https://redis.io/docs/latest/commands/zrangebyscore/) is deprecated. 6.2 itself is EOL.

PR #1036, opened 2025-03-15
fix: remove redundent cache because it cause panels don't show the qu…

DOO-DEV:
in the code if the sync.Map remain, after deletion of each queue in panels like(asynqmon, aysnq-cli,...) the queue doesn't add to the redis set (asynq:queues).
this operation is idempotent. and the cost of it is low

### DIFF
--- a/internal/rdb/rdb.go
+++ b/internal/rdb/rdb.go
@@ -114,12 +114,10 @@ func (r *RDB) Enqueue(ctx context.Context, msg *base.TaskMessage) error {
 	if err != nil {
 		return errors.E(op, errors.Unknown, fmt.Sprintf("cannot encode message: %v", err))
 	}
-	if _, found := r.queuesPublished.Load(msg.Queue); !found {
-		if err := r.client.SAdd(ctx, base.AllQueues, msg.Queue).Err(); err != nil {
-			return errors.E(op, errors.Unknown, &errors.RedisCommandError{Command: "sadd", Err: err})
-		}
-		r.queuesPublished.Store(msg.Queue, true)
+	if err := r.client.SAdd(ctx, base.AllQueues, msg.Queue).Err(); err != nil {
+		return errors.E(op, errors.Unknown, &errors.RedisCommandError{Command: "sadd", Err: err})
 	}
+	r.queuesPublished.Store(msg.Queue, true)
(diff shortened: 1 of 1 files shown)


internal/rdb/rdb.go
```diff
@@ -114,12 +114,10 @@ func (r *RDB) Enqueue(ctx context.Context, msg *base.TaskMessage) error {
 	if err != nil {
 		return errors.E(op, errors.Unknown, fmt.Sprintf("cannot encode message: %v", err))
 	}
-	if _, found := r.queuesPublished.Load(msg.Queue); !found {
-		if err := r.client.SAdd(ctx, base.AllQueues, msg.Queue).Err(); err != nil {
-			return errors.E(op, errors.Unknown, &errors.RedisCommandError{Command: "sadd", Err: err})
-		}
-		r.queuesPublished.Store(msg.Queue, true)
+	if err := r.client.SAdd(ctx, base.AllQueues, msg.Queue).Err(); err != nil {
```
pior, line 117: The point of `r.queuesPublished` is to avoid calling SAdd on every enqueue operation.

1. SAdd is kind of expensive, whether it updates the set or not (measured at ~25% of CPU at high job throughput (>5K/s)
2. Two round-trip to Redis instead of one is a major overhead

If we really need this, the right implementation is to change `r.queuesPublished` for something that would implement a TTL (basically a cache)

kamikazechaser:
The existing sync.Map has negligible overhead even for 100's of queues. If the issue is correctly deleting the queue, the frontend client is expected to use the latest version of the inspector and correctly call DeleteQueue.

And as @pior said, it increases network overhead.

internal/rdb/rdb.go
DOO-DEV: yes we should implement a cache that with a period retrieve the existing queues from the redis and load them into the sync.Map

DOO-DEV:
> The existing sync.Map has negligible overhead even for 100's of queues. If the issue is correctly deleting the queue, the frontend client is expected to use the latest version of the inspector and correctly call DeleteQueue.
> 
> And as @pior said, it increases network overhead.

The issue is the asynq and frontend are separate(seperate sync.Map in separate memory)

PR #1059, opened 2025-06-24
enhancement: allow setting config for each group

armistcxy:
This is not the final version, still need to add test `TestAggregatorMultipleGroupConfigs` to `aggregator_test.go` (though I have test through asynqmon). 

**Concern**: if groupA has `GracePeriod: 2 * time.Hour`, group B has `GracePeriod: 150 * time.Minutes`. Then after 4 hours, tasks from group B can be aggregated, but it supposed to be after 2.5 hours. The reason is: 
```go
interval := defaultAggregationCheckInterval
if params.gracePeriod < interval {
    interval = params.gracePeriod
}
for _, config := range params.groupConfigs {
    interval = min(interval, config.GracePeriod)
}

... 
ticker := time.NewTicker(a.interval)
....
case t := <-ticker.C:
    a.exec(t)
}
```

I will work on this, but before that I need some validations or suggestions from you @kamikazechaser .

### DIFF
--- a/aggregator.go
+++ b/aggregator.go
@@ -31,6 +31,9 @@ type aggregator struct {
 	maxDelay    time.Duration
 	maxSize     int
 
+	// Override group configurations.
+	groupConfigs map[string]GroupConfig
+
 	// User provided group aggregator.
 	ga GroupAggregator
 
@@ -50,6 +53,7 @@ type aggregatorParams struct {
 	maxDelay        time.Duration
 	maxSize         int
 	groupAggregator GroupAggregator
+	groupConfigs    map[string]GroupConfig
 }
 
 const (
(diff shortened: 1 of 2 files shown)


armistcxy:
An example of new asynq server config
```go
	srv := asynq.NewServer(asynq.RedisClientOpt{Addr: "localhost:6379"}, asynq.Config{
		GroupGracePeriod: 10 * time.Second,
		GroupMaxSize:     30,
		GroupMaxDelay:    1 * time.Minute,
		GroupConfigs: map[string]asynq.GroupConfig{
			"group1": {
				GracePeriod: 7 * time.Second,
				MaxSize:     30,
			},
			"group2": {
				GracePeriod: 9 * time.Second,
				MaxSize:     25,
			},
		},
		GroupAggregator: asynq.GroupAggregatorFunc(func(group string, tasks []*asynq.Task) *asynq.Task {
			switch group {
			case "group1":
				return asynq.NewTask("task1", []byte("payload-merged"))
			case "group2":
				payload := make([]byte, 0)
				for _, task := range tasks {
					payload = append(payload, task.Payload()...)
				}
				return asynq.NewTask("task2", payload)
			}
			return nil
		}),
	})
```

armistcxy:
Still need some thoughts/considering from maintainer to decide what to implement next. 

As you can see the problem happens in case the aggregate intervals for all group are smaller than 7 seconds. My example shows that the interval for checking is 3 seconds, this indicates that task from group B will only be send in the next iteration (0 -> 3 -> 6 (this is the moment tasks from group B will be aggregated, but the expected is fourth second)

PR #1070, opened 2025-08-27
Feature: Add Headers Support to Tasks

### LINKED ISSUE #393: [FEATURE REQUEST] Add message headers

joejoe-am:
This MR introduces headers as first-class metadata on Task.
Headers allow attaching cross-cutting metadata (e.g., trace IDs, tenant IDs, feature flags) without mixing them into the business payload.

Motivation
	•	Resolves [Issue #393](https://github.com/hibiken/asynq/issues/393) (support for headers/metadata).
	•	There was an earlier attempt in [PR #838](https://github.com/hibiken/asynq/pull/838), but it became stale and was never merged. This MR picks up the idea, updates it for the current codebase, and ensures backward compatibility.
	
✅ This MR supersedes #838 and #606 and closes #393.


joejoe-am:
Hi Dear @hibiken,
there were many discussions about this feature and I think it's handy feature.
What do you think ?

asynq.go
```diff
@@ -48,6 +52,25 @@ func NewTask(typename string, payload []byte, opts ...Option) *Task {
...
+}
+
+// NewTaskWithHeaders returns a new Task given a type name, payload data, and headers.
+// Options can be passed to configure task processing behavior.
+// TODO: In the next major (breaking) release, fold this functionality into NewTask
+//
+//	so that headers are supported directly. After that, remove this method.
+func NewTaskWithHeaders(typename string, payload []byte, headers map[string]string, opts ...Option) *Task {
+	h := make(map[string]string)
+	for k, v := range headers {
```
kamikazechaser, line 67: Have you looked into maps.Clone as an alternative to this?

joejoe-am: that looks cleaner.
fixed.

joejoe-am:
> Because your change touches a critical path in the processor (retries, failures) I'll wait for a +1 from one of the very large consumers of this library (1k+ tasks/every few sec) to also give their views on this change.
> 
> If possible include some tests as well.

Thanks for the feedback! 🙏

Just to clarify — the changes in this MR don’t introduce any new logic into the processor’s critical path. Headers are simply carried alongside the existing TaskMessage, using the same serialization path as before. If no headers are set, the behavior is exactly the same as today. So there’s no additional branching, no new network round-trips, and no measurable performance impact.

That said, I absolutely agree on the importance of validating this in real scenarios. I’ll add dedicated tests.

yordis:
@kamikazechaser how could we make this to production? Do you need me to help anyhow?

We need it in order to have OTEL integration.

joejoe-am:
Hi @hibiken , @kamikazechaser ,
Just wanted to check in again on this PR. Since headers are a common need and while staying backward compatible, it would be great to know if you plan to review or merge it. Please let me know if there’s anything further I can adjust to help move it forward. 🙏

joejoe-am:
@kamikazechaser @hibiken 
Hi again, this PR has a lot of requests, why isn’t it being followed up, any other issues?

cielu:
Are there any other questions? We're waiting for version 0.26 to be released.

yordis:
@kamikazechaser will you add direct OTEL integration?

armistcxy:
I think it's really cool to add an option that allow integrating OTEL into asynq seamlessly

PR #1073, opened 2025-09-19
Add --tls option to dash command

thomas4019:
The current TLS options are insufficient to be able to easily connect to an AWS Redis instance with TLS enabled. In this PR, I'm adding a --tls argument which works the same way as the redis-cli, when specified it connects over TLS, but doesn't specify any of other TLS settings.

### DIFF
--- a/tools/asynq/cmd/root.go
+++ b/tools/asynq/cmd/root.go
@@ -41,6 +41,7 @@ var (
 	clusterAddrs    string
 	tlsServerName   string
 	insecure        bool
+	useTLS          bool
 )
 
 // rootCmd represents the base command when called without any subcommands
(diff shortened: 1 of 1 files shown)


tools/asynq/cmd/root.go
```diff
@@ -402,11 +405,16 @@ func getRedisConnOpt() asynq.RedisConnOpt {
 }
 
 func getTLSConfig() *tls.Config {
+	if viper.GetBool("tls") {
+		return &tls.Config{InsecureSkipVerify: viper.GetBool("insecure")}
```
kamikazechaser: Looks a bit confusing to users who might want to use both and then it returns early on --tls and then it can't connect in instances where the hostname doesn't match.

PR #1083, opened 2025-10-30
feat: done add-username-cli

vlle:
I had a problem with asynq CLI due to no option to pass username in any way possible. Now it's passable via -U flag. I added it to both code and README.md example


kamikazechaser:
LGTM.

PR #1091, opened 2026-02-06
支持配置LeaseDuration、HeartbeatInterval

Lcys:
支持配置LeaseDuration、HeartbeatInterval参数，减少lease expired 错误


kamikazechaser:
See GH actions logs. You need to update `internal/rdb/rdb_test.go` and `processor_test.go`.

PR #1094, opened 2026-02-24
Add BatchEnqueue for pipelined multi-task enqueue

enilsen16:
Adds BatchEnqueue to the Broker interface and RDB implementation that sends multiple enqueueCmd Lua script invocations in a single Redis pipeline round-trip. Also adds BatchEnqueueContext to the Client as the public API, returning per-task results for partial-success handling.

Ref: hibiken/asynq#1069

### DIFF
--- a/internal/rdb/rdb.go
+++ b/internal/rdb/rdb.go
@@ -143,15 +143,14 @@ func (r *RDB) Enqueue(ctx context.Context, msg *base.TaskMessage) error {
 // Each item is either enqueued immediately (ProcessAt is zero) or added to the
 // scheduled sorted set.
 //
+// WARNING: tasks whose IDs already exist in Redis are silently skipped.
+//
 // The pipeline executes independent Lua scripts per task — there is no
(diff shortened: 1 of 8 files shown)


kamikazechaser:
Please document any atomicity guarantees e.g. what kind of failures would or would not reject the entire batch.

internal/rdb/rdb.go
```diff
@@ -139,6 +139,76 @@ func (r *RDB) Enqueue(ctx context.Context, msg *base.TaskMessage) error {
 	return nil
 }
 
+// BatchEnqueue adds all given tasks to their respective pending lists using a
+// single Redis pipeline round-trip. It returns the number of newly enqueued
+// messages (tasks whose IDs already exist in Redis are silently skipped).
+// BatchEnqueue adds all given tasks to Redis using a single pipeline round-trip.
+// Each item is either enqueued immediately or scheduled based on its ProcessAt field.
```
kamikazechaser: Add a clear warning on silent skipping of duplicates on its own line e.g.

`WARNING: tasks whose IDs already exist in Redis are silently skipped`

PR #1096, opened 2026-03-17
feat: Add an option to specify headers

kayoch1n:
This PR adds `Header()` option to the Client API, allowing users to associate  key-value headers to tasks.
- Add `Header(key, value string) Option` function to client.go.
- Add parsing support in inspector.go. 
- Add tests.

The underlying value of `Header` is `[2]string`. Inspector parses Header option by unmarshaling the JSON array argument back to `[2]string`

usage:
```go
client.Enqueue(NewTask("send_email", data), Header("channel", "email"))
```

Follow-up to #1070 which added headers support to Task messages

### DIFF
--- a/client.go
+++ b/client.go
@@ -234,6 +234,13 @@ func (h headerOption) String() string {
 func (h headerOption) Type() OptionType   { return HeaderOpt }
 func (h headerOption) Value() interface{} { return [2]string{h[0], h[1]} }
 
+func Headers(headers map[string]string) (ret []Option) {
+	for key, value := range headers {
+		ret = append(ret, Header(key, value))
+	}
+	return
+}
+
 // ErrDuplicateTask indicates that the given task could not be enqueued since it's a duplicate of another task.
 //
 // ErrDuplicateTask error only applies to tasks enqueued with a Unique option.
--- a/client.go
+++ b/client.go
@@ -6,7 +6,9 @@ package asynq
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"maps"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


client.go
```diff
@@ -217,6 +221,26 @@ func (name groupOption) String() string     { return fmt.Sprintf("Group(%q)", st
...
+
+func (h headerOption) String() string {
+	var bytes []byte
+	bytes, _ = json.Marshal(h)
+	return fmt.Sprintf("Header(%s)", bytes)
+}
+func (h headerOption) Type() OptionType   { return HeaderOpt }
+func (h headerOption) Value() interface{} { return [2]string{h[0], h[1]} }
+
+func Headers(headers map[string]string) (ret []Option) {
```
kamikazechaser: Could we rename this to something else like HeadersMap so that it is not confusing. #1070 also introduced a Headers method on Task.

kamikazechaser:
Maybe another thing to document is when to use this vs what was introduced in #1070.

client.go
kayoch1n: 👌 I have removed this function.

kayoch1n:
> Maybe another thing to document is when to use this vs what was introduced in #1070.

I've added documentation to the Header function explaining when to use it vs `NewTaskWithHeaders`:

  - Use Header() Option when composing headers with other enqueue-time options (e.g., MaxRetry, Queue)
  - Use NewTaskWithHeaders when headers are an intrinsic part of the task definition

Besides, the `Header` option follows the same pattern as other options (implements String(), Type(), Value()), ensuring compatibility with existing infrastructure like scheduler serialization.

Issue #669, opened 2023-06-12
No option to specify DB for Redis Sentinel URI


importnil:
I've come to a problem with `asynqmon` that I couldn't get any details from Redis with databases other than 0 when connecting through Sentinels. Turns out that the final `ParseRedisURI` function doesn't extract DB parameter value, e.g.:

```
redis-sentinel://:rfs-service:26379/3?master=mymaster
```

Refs:
https://github.com/hibiken/asynqmon/blob/5c48e4e31d6bbfe8f5ee0142ee1bc563aefaa558/cmd/asynqmon/main.go#LL106C32-L106C32
https://github.com/hibiken/asynq/blob/fde294be326a252f6fcb1d942c11c055bebeecfb/asynq.go#LL522C9-L522C9

If it's not intended behaviour, would it make sense to you that I file a PR for this feature?

PR #1100, opened 2026-03-21
fix: parse DB number from redis-sentinel URI path

ljluestc:
fix: parse DB number from redis-sentinel URI path (#669)

Problem

parseRedisSentinelURI ignored the /dbnumber path segment, making it impossible to connect to any DB other than 0 via sentinel URIs. For example:
text
would silently drop the /3 and always use DB 0.

The redis:// and rediss:// schemes already supported this via parseRedisURI.

Changes

•  asynq.go: Extract the DB number from the URI path in parseRedisSentinelURI, using the same strconv.Atoi(xs[0]) pattern as parseRedisURI. Pass the parsed DB into RedisFailoverClientOpt. Return a descriptive error if the path segment is not a valid integer.
•  asynq.go doc comment: Updated the ParseRedisURI godoc to show /dbnumber in the sentinel URI format:
text
•  asynq_test.go: Added test cases:
◦  redis-sentinel://...5002/3?master=mymaster → DB 3
◦  redis-sentinel://:mypassword@...5002/7?master=mymaster → DB 7 with password
◦  redis-sentinel://localhost:5000/abc?master=mymaster → error (non-integer)

### DIFF
--- a/asynq.go
+++ b/asynq.go
@@ -471,7 +471,7 @@ func (opt RedisClusterClientOpt) MakeRedisClient() interface{} {
 //	redis://[:password@]host[:port][/dbnumber]
 //	rediss://[:password@]host[:port][/dbnumber]
 //	redis-socket://[:password@]path[?db=dbnumber]
-//	redis-sentinel://[:password@]host1[:port][,host2:[:port]][,hostN:[:port]][?master=masterName]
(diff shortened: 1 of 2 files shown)


kamikazechaser (approved):

kamikazechaser:
@ljluestc Let me know if you want to add something else to it cause it was marked as draft.

PR #1101, opened 2026-04-04
refactor(cli): migrate commands from Run to RunE and fix group error …

sevico:
…handling

Migrate all CLI command handlers from cobra's Run to RunE, replacing fmt.Println(err) + os.Exit(1) patterns with idiomatic error returns. This improves testability and lets cobra handle errors consistently via the existing SilenceUsage/SilenceErrors configuration.

Also refactor getDuration() and getTime() helpers to return errors instead of calling os.Exit(1).

Fix a bug in group.go where the error from inspector.Groups() was never checked, causing silent failures (e.g. on Redis connection errors) to be misreported as "No groups found".

### DIFF
--- a/tools/asynq/cmd/queue.go
+++ b/tools/asynq/cmd/queue.go
@@ -7,7 +7,6 @@ package cmd
 import (
 	"fmt"
 	"io"
-	"os"
 
 	"github.com/MakeNowJust/heredoc/v2"
 	"github.com/fatih/color"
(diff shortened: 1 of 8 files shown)


tools/asynq/cmd/queue.go
```diff
@@ -233,49 +231,63 @@ func printDailyStats(stats []*asynq.DailyStats) {
...
-func queueUnpause(cmd *cobra.Command, args []string) {
+func queueUnpause(cmd *cobra.Command, args []string) error {
 	inspector := createInspector()
+	var firstErr error
 	for _, qname := range args {
 		err := inspector.UnpauseQueue(qname)
 		if err != nil {
 			fmt.Println(err)
+			if firstErr == nil {
+				firstErr = err
```
kamikazechaser, line 259: Generally first error checking is fine and there is no need for aggregation, but this might be a use case for errors.Join because the caller may want to know if there were other UnpauseQueue errors.

sevico: In this context, the caller is cobra's RunE — it only prints the error and sets the exit code. There's no upstream code inspecting individual errors with errors.Is/errors.As.
Each error is already printed inline during the loop via fmt.Println(err), so the user sees all failures as they happen. If we switch to errors.Join, cobra would print the joined error again at the end,duplicating the output.
I think firstErr is sufficient here — it signals "something went wrong" for the exit code, while individual errors are already surfaced inline. Happy to reconsider if there's a caller I'm missing that would benefit from the aggregated error.

PR #1102, opened 2026-04-07
fix: allow re-enqueue of archived/completed tasks with same TaskID

### LINKED ISSUE #842: [QUESTION] Archived periodic tasks are not triggered on the scheduled run
**Describe the bug**

When following the advise given here: https://github.com/hibiken/asynq/discussions/376 to specify a fixed task ID for a periodic task to avoid duplicate tasks for multiple scheduler instances, for example:

```golang
task := asynq.NewTask(name, payload,  asynq.TaskID("foo"), asynq.Retention(time.Minute))
_, err := scheduler.Register("* * * * *", task)
```

periodic tasks that error out are simply moved to the `archived` state and are never triggered again on the next schedule.

I can confirm this doesn't happen if the `TaskID()` option is not specified. The only way I've managed to get the task to start again on the next schedule is to manually delete the task (remove it from archived so it can be re-enqueued again).

We write our background tasks to be completely idempotent so simply running a failed task on the next scheduled run is perfectly fine for us. 

Is there a way to skip the failed periodic tasks from going to `archived`? It would be ideal if an option similar to `Retention()` exists (which as I understand only works for `completed` tasks) for archived tasks.

kevinsanz93:
## Summary

Fixes #842

When using a fixed `TaskID` for periodic tasks (as recommended in #376 to avoid duplicates across multiple scheduler instances), tasks that fail and move to the `archived` state are never re-enqueued on the next schedule. The same applies to `completed` tasks with `Retention`.

**Root cause:** All six enqueue-family Lua scripts (`enqueueCmd`, `enqueueUniqueCmd`, `scheduleCmd`, `scheduleUniqueCmd`, `addToGroupCmd`, `addToGroupUniqueCmd`) use a blind `EXISTS` check on the task hash key. Since archived/completed tasks keep their hash key in Redis, `EXISTS` returns 1 and the script returns a `TaskIDConflict` error, permanently blocking the periodic task.

**Fix:** Each script now checks the task's `state` field when the key exists. If the state is `archived` or `completed`, the stale entry is atomically cleaned up (removed from the sorted set, hash deleted) and the enqueue proceeds. Active states (`pending`, `active`, `scheduled`, `retry`, `aggregating`) still return a conflict as before.

A shared Lua fragment (`clearStaleLua`) and Go helper (`enqueueBaseKeys`) keep the logic DRY across all six scripts.

## Changes

- `internal/rdb/rdb.go` -- Added `clearStaleLua` function (generates reusable Lua snippet) and `enqueueBaseKeys` helper. Updated all 6 Lua scripts and their Go callers.
- `internal/rdb/rdb_test.go` -- Added 4 new tests:
  - `TestEnqueueOverwritesArchivedTaskWithSameID`
  - `TestEnqueueOverwritesCompletedTaskWithSameID`
  - `TestScheduleOverwritesArchivedTaskWithSameID`
  - `TestAddToGroupOverwritesArchivedTaskWithSameID`

## Test plan

- [x] New tests confirm archived/completed tasks are overwritten on re-enqueue
- [x] Existing `TestEnqueueTaskIdConflictError`, `TestScheduleTaskIdConflictError`, `TestAddToGroupeTaskIdConflictError` still pass (in-flight state conflicts preserved)
- [x] Full `internal/rdb` test suite passes (103 tests)
- [x] Full project test suite passes (`go test ./...`)
- [ ] Tests should also be run with `--redis_cluster` flag per CONTRIBUTING.md

### DIFF
--- a/internal/rdb/rdb.go
+++ b/internal/rdb/rdb.go
@@ -82,11 +82,51 @@ func (r *RDB) runScriptWithErrorCode(ctx context.Context, op errors.Op, script *
 	return n, nil
 }
 
+// clearStaleLua returns a Lua fragment that checks whether a task key already
+// exists. If the task is in "archived" or "completed" state, it removes the
+// stale entry from the corresponding sorted set and deletes the hash so the
+// caller can proceed to enqueue. For any other state it returns 0 (conflict).
+//
+// Parameters are 1-based Lua KEYS/ARGV indices:
+//   - taskKeyIdx:      KEYS index of the task hash key
+//   - archivedKeyIdx:  KEYS index of the archived sorted set key
+//   - completedKeyIdx: KEYS index of the completed sorted set key
+//   - taskIDArgvIdx:   ARGV index of the task ID
+func clearStaleLua(taskKeyIdx, archivedKeyIdx, completedKeyIdx, taskIDArgvIdx int) string {
+	return fmt.Sprintf(`
+if redis.call("EXISTS", KEYS[%d]) == 1 then
+	local state = redis.call("HGET", KEYS[%d], "state")
+	if state == "archived" then
+		redis.call("ZREM", KEYS[%d], ARGV[%d])
+		redis.call("DEL", KEYS[%d])
+	elseif state == "completed" then
+		redis.call("ZREM", KEYS[%d], ARGV[%d])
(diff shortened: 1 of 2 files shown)


kevinsanz93:
I do believe this is ready for review

Issue #173, opened 2020-06-16
[FEATURE REQUEST] Support RabbitMQ queue 


toszo:
**Describe the solution you'd like**
RabbitMQ as a message broker for tasks. 

**Describe alternatives you've considered**
MQTT broker which would enable IoT usage. 

**Additional context**
Working on a distributed system that part of which will run on the Edge devices.

hibiken:
@toszo Thank you for opening this issue! 

I've thought about adding support for other message brokers and RabbitMQ is one of them. And this is definitely in the project's roadmap. However, I'd like to work on this after we have solid features implemented with Redis as a broker. If you need support for RabbitMQ soon, I'd say it's better to look for other solutions. It'll probably take a few months before I start working on adding support for other message brokers :)

zhaolion:
😄 Hello, @hibiken.  I had use Kafka & Redis as message broker for async tasks system at the same time(an ugly combination, but it's useful), I hope you will consider adding support for Kafka/RabbitMQ :)

zhaolion:
> @toszo Thank you for opening this issue!
> 
> I've thought about adding support for other message brokers and RabbitMQ is one of them. And this is definitely in the project's roadmap. However, I'd like to work on this after we have solid features implemented with Redis as a broker. If you need support for RabbitMQ soon, I'd say it's better to look for other solutions. It'll probably take a few months before I start working on adding support for other message brokers :)

How can I get the solid features plan in your design, Maybe I can contribute some features 🤓

hibiken:
@zhaolion thanks for commenting!
I've been exploring Kafka recently (I don't have any experience with it, so I need to learn it first 😄 ). I think it'd be great if `asynq` can also support Kafka as a message broker.

In terms of features to implement before working on adding support for other message brokers, I'd like to at least resolve these issues:

- Periodic Task feature #183 
- Support Redis Cluster #182

Once I have tackled those issues, I think we can start looking into adding support to Kafka, RabbitMQ, etc

rugwirobaker:
Nats would be nice too. The only reason I hesitate with Redis is data persistence with Redis is a bit more involved. I would love to help with nats of course if there is an isolated Store interface.

leafmirror:
maybe rocketmq is better

Me1onRind:
An idea, other message queues are supported as plugin. Use interface to decoupling from specific storage.

iqDF:
[...]

PR #1122, opened 2026-04-14
Decouple PubSub from Redis in the Broker interface

halomanlodestar:
A sub part of #173

## Summary
This change addresses the TODO on `CancelationPubSub()` in Broker interface at `internal/base/base.go:726`.
Removes the hard dependency on `*redis.PubSub` from the Broker interface, replacing it with a minimal `PubSub` interface. This is a prerequisite for supporting non-Redis broker implementations.

# Motivation
`CancelationPubSub()` in the `Broker` interface previously returned `*redis.PubSub` directly, making it impossible to implement a compliant broker without Redis. This change abstracts that dependency behind an interface.

### New `base.PubSub` interface

- Defines a broker-agnostic `PubSub` interface with `Channel() <-chan string`, and `Close`
- `Broker.CancelationPubSub()` now returns (PubSub, error) instead of (*redis.PubSub, error)

### New RedisPubSub adapter
- Wraps `*redis.PubSub` to implement `base.PubSub`
- Channel() returns <-chan string (message payloads only), hiding Redis-specific message types from callers
- Updated consumers

## Testing
Existing `TestCancelationPubSub` updated to match the new <-chan string API (reads msg directly instead of msg.Payload).

### DIFF
--- a/client.go
+++ b/client.go
@@ -44,7 +44,6 @@ func NewClient(r RedisConnOpt) *Client {
 // NewClientFromRedisClient returns a new instance of Client given a redis.UniversalClient
 // Warning: The underlying redis connection pool will not be closed by Asynq, you are responsible for closing it.
 func NewClientFromRedisClient(c redis.UniversalClient) *Client {
-	// redis.PubSub
 	return &Client{broker: rdb.NewRDB(c), sharedConnection: true}
 }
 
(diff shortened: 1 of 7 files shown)


kamikazechaser:
This is out of scope. See #1138.

PR #1127, opened 2026-04-29
Try disabling memory usage profiling

nycdavid:


kamikazechaser:
Add inline docs, documenting this feature.

kamikazechaser (approved):

PR #1133, opened 2026-05-14
fix: add panic recovery to goroutines executing user-provided callbacks

### LINKED ISSUE #1132: Panic in user-provided callbacks (HealthCheckFunc, GroupAggregator, PeriodicTaskConfigProvider) crashes worker process
## Problem

Three user-provided callbacks are called inside library goroutines without `defer recover()`. A panic in any of them kills the entire worker process, terminating all in-flight tasks.

## Affected callbacks

| Callback | File | Called in goroutine at |
|----------|------|-----------------------|
| `HealthCheckFunc` | healthcheck.go:75 | healthcheck.go:64 |
| `GroupAggregator.Aggregate` | aggregator.go:161 | aggregator.go:122 (`go a.aggregate(t)`) |
| `PeriodicTaskConfigProvider.GetConfigs` | periodic_task_manager.go:209 | periodic_task_manager.go:131 |

## Example: common HealthCheckFunc bug

```go
srv := asynq.NewServer(redisOpt, asynq.Config{
    HealthCheckFunc: func(err error) {
        // Bug: err is nil when ping succeeds
        metrics.Record("health", err.Error()) // PANIC: nil pointer dereference
    },
})
```

The library calls `HealthCheckFunc(nil)` when the broker ping succeeds. The user's code dereferences `err.Error()` on nil. The goroutine panics. The entire worker process dies. All in-flight tasks are lost.

## Why this matters

`processor.perform()` already recovers from panics in task handlers (processor.go:424). The same principle applies to these other user-facing interfaces: the library spawns the goroutine, the library calls the user's code, so the library is responsible for recovering if that code panics.

blackwell-systems:
Fixes #1132.

## Problem

Three user-provided callbacks are called inside library goroutines without `defer recover()`. A panic in any of them kills the entire worker process, terminating all in-flight tasks.

`processor.perform()` already has this pattern for task handlers (processor.go:424). These three callbacks were missed.

## Fix

Add `defer recover()` with stack trace logging to:

| File | Callback | What happens on panic |
|------|----------|----------------------|
| healthcheck.go | `HealthCheckFunc(error)` | Log stack trace, goroutine exits (health checks stop) |
| aggregator.go | `GroupAggregator.Aggregate(group, tasks)` | Log stack trace, aggregation check exits (semaphore released) |
| periodic_task_manager.go | `PeriodicTaskConfigProvider.GetConfigs()` | Log stack trace, syncer goroutine exits |

Recovery style matches the existing pattern in `perform()`: full `debug.Stack()` output via the component's logger.

## Tests

- `TestHealthChecker_PanicRecovery`: HealthCheckFunc panics on every call, process survives
- `TestHealthChecker_PanicRecovery_NilError`: common bug where user calls `err.Error()` on nil (ping succeeded)
- `TestAggregator_PanicRecovery`: GroupAggregator panics, process survives


kamikazechaser:
> user-provided callbacks

All 3 are the user's responsibility to implement correctly.

blackwell-systems:
I agree users should implement callbacks correctly. But the library already recovers from panics in task handlers (processor.go:424) for exactly this reason: one bad handler shouldn't crash the entire worker and kill all in-flight tasks.

These three callbacks run in the same kind of library-owned goroutine. The risk is the same: a panic in a health check shouldn't terminate unrelated task processing. The fix is consistent with the existing recovery pattern, not a new philosophy.

Happy to close if the maintainers prefer a different approach.

Issue #1016, opened 2025-02-10
[FEATURE REQUEST] Logger interface support `context.Context`


poetlife:
**Is your feature request related to a problem? Please describe.**

Current `log.Base` interface doesn't support explicitly pass `ctx` param which is important for log observability such as OpenTelemetry. 

**Describe the solution you'd like**

Provide another interface which accepts a `ctx context.Context` parameter, such as:
```go
type LoggerWithContext interface {
	// Debug logs a message at Debug level.
	Debug(ctx context.Context, args ...interface{})

	// ...
}
```

**Describe alternatives you've considered**

No alternatives.

**Additional context**

In production environments, we need something like `trace_id` to relate many logs. So, it's important to log with context.

jittery-droid:
My company also needs this feature

gmogan-do-student:
Its currently using https://pkg.go.dev/log#Logger interface

I'm more than happy to migrate things to slog if @hibiken is interested, though I'm not sure how easy it'll be able to be done that doesn't break everyones existing implementations.

halkeye:
my bad, used the test account for ☝️ message

#1137 would be a start for supporting log context

PR #1137, opened 2026-05-19
Augment logger to be structured

halkeye:
Abused work's cursor account to to the heavy lifting updating everything

* Add logging wrapper to make existing logger speak slog
* Update log statements to include context and have fields separated

As far as I can tell, this shouldn't break any existing logging other than modifying the format slightly, by dropping the field=value type thing
Also cursor added component everywhere, which i figured wasn't a bad idea.

still would need some work so proper context gets used/passed through everywhere to fully support https://github.com/hibiken/asynq/issues/1016 but its a start.

### DIFF
--- a/aggregator.go
+++ b/aggregator.go
@@ -85,7 +85,7 @@ func (a *aggregator) shutdown() {
 	if a.ga == nil {
 		return
 	}
-	a.logger.Debug("Aggregator shutting down...")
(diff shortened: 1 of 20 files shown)


halkeye:
@kamikazechaser when you get a minute can you re-enable workflows? I had llm add more test coverage so codecov would be happier

PR #1138, opened 2026-05-23
Add pluggable backend foundation

darwvin-dev:
### What does this PR do?

Adds a small pluggable backend foundation so Asynq clients and servers can be backed by a storage implementation that is not tied directly to Redis internals.

This introduces:

- A public `Backend` interface for enqueue/dequeue/task lifecycle operations.
- A public `TaskMessage` storage-facing representation.
- `NewClientWithBackend` for backend-backed enqueue flows.
- `NewServerWithBackend` and an internal backend runner for backend-backed processing.
- Redis behavior preserved as the existing default path.

### Motivation

Asynq currently exposes Redis connection options as the only first-class backend path. This PR creates the minimum public surface needed to experiment with alternate storage backends while keeping the existing Redis APIs and behavior unchanged.

No linked issue currently exists for this foundation PR.

### Design notes

- The existing Redis-backed `NewClient`, `NewServer`, and `NewServerFromRedisClient` paths remain unchanged by default.
- Backend-backed clients route enqueue/schedule operations through `Backend.Enqueue`.
- Backend-backed servers use a separate internal runner instead of changing the Redis processor pipeline.
- Backend implementations can return `ErrFeatureNotSupported` for behavior that is not part of the minimum contract, such as grouped/unique Redis-specific behavior.
- The `TaskMessage` type intentionally mirrors the storage-facing fields needed by the backend contract rather than exposing internal Redis message types directly.

### Compatibility

This is intended to be additive. Existing Redis users should continue to use the same constructors and code paths as before.

### Tests

I do not have a completed CI run attached to this PR yet. The change is structured so the Redis default path remains the existing behavior, and the new backend path is isolated behind the new constructors.

### Review focus

The main thing I would like feedback on is the shape of the initial `Backend` interface and whether this is the right minimum foundation before adding concrete backend implementations.

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -1,0 +1,77 @@
+// Copyright 2020 Kentaro Hibino. All rights reserved.
+// Use of this source code is governed by a MIT license
+// that can be found in the LICENSE file.
+
+package asynq
+
+import (
+	"context"
+	"errors"
+	"time"
+
+	"github.com/hibiken/asynq/internal/base"
+)
+
+// ErrFeatureNotSupported indicates that a backend does not support the requested feature.
+var ErrFeatureNotSupported = errors.New("asynq: feature is not supported by backend")
+
+// ErrNoProcessableTask indicates that a backend has no task ready for processing.
+var ErrNoProcessableTask = errors.New("asynq: no processable task")
+
+// TaskMessage is the storage-facing representation used by pluggable backends.
+type TaskMessage struct {
+	ID           string
+	Type         string
+	Payload      []byte
+	Headers      map[string]string
+	Queue        string
+	Retry        int
+	Retried      int
+	ErrorMsg     string
+	LastFailedAt time.Time
+	Timeout      time.Duration
+	Deadline     time.Time
+	UniqueKey    string
+	GroupKey     string
+	Retention    time.Duration
+	CompletedAt  time.Time
+	RunAt        time.Time
+}
+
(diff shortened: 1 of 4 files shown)


kamikazechaser:
Yeah I'm not sure about this. It is better to use other libraries that make it easy to plug any backend. This is heavily coupled with Redis.

darwvin-dev:
Thanks, that concern makes sense.

To clarify the goal: I do not want to make the main Asynq package heavier or pull every possible storage dependency into core. The direction I have in mind is the opposite: keep Asynq core small and Redis-first, but define a narrow storage/backend contract so other backends can be provided as optional adapters/plugins.

Ideally, the core package would only own the stable task lifecycle abstraction:

- enqueue / schedule
- lease or dequeue ready tasks
- ack completed tasks
- retry / archive failed tasks
- ping / close
- capability reporting for optional features

Concrete backends such as Postgres, MySQL, Kafka, NATS, etc. should live outside the core package or in separate optional modules, so users only pay for the backend they import.

I agree the current PR still mirrors Redis internals too much. Maybe this should start as an RFC/design issue instead of a full implementation PR. The main question I wanted to validate is whether maintainers are open to a lightweight backend/plugin abstraction in principle, with Redis remaining the default implementation.

If this direction is outside the scope of Asynq, I understand and can close the PR. If it is worth exploring, I can pause this PR and open a design proposal focused on the abstraction boundary and keeping core dependencies minimal.

PR #1149, opened 2026-06-15
Close inspector instances intentionally in asynq cli tool

DeamonDev:
Currently in various `cobra` commands we spawn `asynq.Inspector`'s which are not intentionally closed. In that context it is not huge deal since after command completion, the resources are cleaned up thanks to very nature of `cobra` being used. 

This can also be useful if someone, hypothetically in the future, wants to use some piece of code from these commands elsewhere in the repository (as long as unshared redis connection being used).


DeamonDev:
@hibiken

PR #1152, opened 2026-06-17
Feat: Group-based Sequential Processing (Serialization by Key)

### LINKED ISSUE #1090: [Feature Request] Support for Group-based Sequential Processing (Serialization by Key)
I'm facing a race condition issue when processing tasks that share a specific resource constraint (e.g., tasks affecting the same `wallet_address` and `chain_id`).

Currently, `asynq` allows controlling global concurrency, but it doesn't seem to support "Partitioned Concurrency"—where tasks are executed sequentially *per entity* (same key) but concurrently *across different entities*.

For example:
- Task A1 and Task A2 (both for User A) are enqueued.
- Task B1 (for User B) is enqueued.
- I want Task B1 to run in parallel with Task A1.
- However, Task A2 **must wait** until Task A1 completes.

Currently, if multiple workers pick up Task A1 and Task A2 simultaneously, we encounter state conflicts.

I would like to propose adding a `Group` or `SerializationKey` option when enqueuing a task.
The scheduler/processor should ensure that for any given `Key`, only **one** task is being processed at a time across all workers.

knightfall22:
### Implementation of Key Based Task Synchronization as described in https://github.com/hibiken/asynq/issues/1090.
- Uses `SynchronizationKey` to group tasks conceptually
- A worker holds a lock on a group for the duration of its process time
- Locks have a ttl of the default lease duration(30s), a lock is renewed for the processing duration 

Closes https://github.com/hibiken/asynq/issues/1090


knightfall22:
@hibiken @kamikazechaser

PR #1154, opened 2026-06-21
Add a runnable basic example

DeamonDev:
Adds a minimal, end-to-end example under `_examples/basic` so newcomers can run the quickstart code instead of just reading it. Beyond onboarding, since the example lives inside the module and imports the local package directly, it doubles as a handy entry point for stepping through and debugging `asynq`'s internals with real tasks flowing through.


DeamonDev:
@hibiken

PR #1158, opened 2026-06-29
fix: recover orphaned active tasks with no lease entry

### LINKED ISSUE #1144: [BUG] Orphaned Tasks Stuck in ACTIVE State Without Lease Data (Unable to Retry)
**Describe the bug**
In our production environment, we have observed a state consistency issue where tasks are marked as ACTIVE but their corresponding lease records are missing from the queue.
These tasks appear to be "orphaned"—they hold an active status indefinitely but are never picked up by workers or the scheduler for re-execution. This leads to a permanent deadlock for these specific tasks.

**Expected behavior**
If a task is ACTIVEbut its lease expires or is lost, the system should:
Detect the inconsistency.
Automatically transition the task back to a PENDINGstate.
Allow it to be rescheduled and retried by available workers.

**Actual Behavior**
The scheduler seems unable to detect these "orphaned" tasks. Since they are technically still ACTIVE, they are not picked up for retry logic, effectively hanging forever.

**Additional**
During the service restart, it will be removed from the whitelist and lose access to Redis (existing connections may persist briefly while new connections will fail).

DeamonDev:
Tasks could become stuck in the active list with no matching entry in the lease `ZSET` (e.g. lease-key eviction under maxmemory, a Redis restore/failover, or external key mutation). The recoverer only scans the lease `ZSET`, so these orphans were never retried or archived and stayed active forever.

The recoverer now re-attaches an already-expired lease to any active task that is missing one, so the existing lease-expiry recovery path picks them up and routes them through retry/archive in the same pass.

Closes #1144


DeamonDev:
@hibiken

PR #1171, opened 2026-07-29
feat: make archive retention configurable to prevent silent task loss

nauanelinhares:
## Summary

Asynq keeps failed tasks in a per-queue archive set so operators can inspect and debug them. Today that archive is bounded by two hardcoded limits:

- **Max size:** 10,000 tasks per queue
- **Retention:** 90 days

When a service processes a large volume of tasks and many of them eventually fail (after exhausting retries), the archive fills up quickly. Once the cap is reached, **the oldest archived tasks are permanently deleted** to make room for new ones. There is no warning, no metric, and no way to opt out — the data is simply gone.

This PR makes archive retention configurable via global `Server` config:

```go
srv := asynq.NewServer(redisOpt, asynq.Config{
    MaxArchiveSize:           50000, // default: 10000
    ArchivedExpirationInDays: 180,   // default: 90
})
```

It also fixes an off-by-one bug in the archive trim logic: the Lua scripts used `ZRANGE` / `ZREMRANGEBYRANK` with `0, -N`, which kept **N-1** tasks instead of **N**. With the previous hardcoded limit of 10,000 this was invisible, but once the size becomes configurable, `MaxArchiveSize: 1` would drop the entire archive — including the task that was just archived.

### Changes

- Add `Config.MaxArchiveSize` and `Config.ArchivedExpirationInDays` on the server (global, not per-queue)
- Move archive limits from package-level constants to per-`RDB` instance fields via `RDBConfig`
- **Server** persists settings to Redis (`asynq:archive_config`) once on `Start()`
- **Inspector** loads settings from Redis once on creation; trim uses in-memory config only (no Redis read per archive)
- Add optional `InspectorConfig` to override defaults explicitly when needed
- Fix off-by-one trim in all five archive scripts (`Archive`, `ArchiveTask`, `ArchiveAllPendingTasks`, `ArchiveAllAggregatingTasks`, `archiveAll`)
- Add regression tests, including `MaxArchiveSize: 1`

### How config propagates

| Component | Source of truth |
|-----------|-----------------|
| Server (processor, recoverer) | `Config` passed to `NewServer` |
| Inspector / asynqmon | Loads from Redis on startup (written by Server), or `InspectorConfig` override |
| Client / Scheduler / CLI | Defaults unless given `RDBConfig` |

**Note:** Inspector/asynqmon only see Server settings after the Server has started and written to Redis. If Inspector starts before Server, it uses defaults until restarted (or until `InspectorConfig` is provided).

Per-queue archive config was explored and intentionally left out to keep this change focused.

## Motivation

| Scenario | Before | After |
|----------|--------|-------|
| High-failure batch job archives 15k tasks/day | Only last 10k kept; 5k+ lost daily | Operator sets `MaxArchiveSize: 100000` |
| Compliance requires 6-month failure history | Impossible (90-day hard limit) | Operator sets `ArchivedExpirationInDays: 180` |
| Small dev environment | 10k default wastes Redis memory | Operator lowers both values |
| `MaxArchiveSize: 1` | Archive wiped on every archive (off-by-one) | Exactly 1 task kept |
| asynqmon manual archive/delete | Always used 10k/90 defaults | Picks up Server settings from Redis |

## Test plan

- [x] `TestNewRDBArchiveConfig` — defaults and normalization
- [x] `TestLoadArchiveConfigFromRedis` — Inspector loads Server settings from Redis
- [x] `TestArchiveWithCustomMaxArchiveSize` — trim respects custom limit via `Archive`
- [x] `TestArchiveWithMaxArchiveSizeOfOne` — regression for off-by-one (`MaxArchiveSize: 1`)
- [x] `TestArchiveWithCustomArchivedExpirationInDays` — expired tasks purged per custom TTL
- [x] `TestArchiveAllPendingTasksTrimsToMaxArchiveSize` — trim via Inspector path Lua script
- [x] `TestArchiveTrim` updated — expects exactly 10,000 tasks, not 9,999
- [x] Full suite: `go test -race ./...`
- [x] Manual: start server with custom limits, confirm asynqmon/Inspector respect them after Server `Start()`

image


nauanelinhares:
@kamikazechaser can you see? 🙏🏾

ns-skaranth:
@kamikazechaser would you be able to review and merge this change?

